PR #57, opened 2024-08-05
Conn

rustaceanrob:


src/peers/traits.rs
```diff
@@ -45,31 +45,30 @@ pub(crate) trait MessageGenerator {
...
+    fn connect(&mut self, addr: AddrV2, port: u16) -> FutureResult<(Reader, Writer), PeerError>;
 }
 
 pub(crate) struct ClearNetConnection {}
 
+impl ClearNetConnection {
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+}
```
oleonardolima, line 59: @rustaceanrob Shouldn't this be the `ConnectionType::ClearNet` variant instead ?

rustaceanrob: Yeah good point. The trait can be implemented over the enum. Will refactor later

PR #59, opened 2024-08-06
node: add tor support

rustaceanrob:


src/lib.rs
```diff
@@ -18,6 +18,13 @@ mod prelude;
 
 use std::net::IpAddr;
 
+#[cfg(feature = "tor")]
+pub use arti_client::TorClient;
+#[cfg(feature = "tor")]
+pub use arti_client::TorClientConfig;
```
oleonardolima, line 24: I think you can fit this under a single use, can't you? 🤔

oleonardolima (commented):
It's looking good! I'll take a look on pinning the dependencies.

Maybe we should also have a separate example just for Tor usage 🤔

src/lib.rs
rustaceanrob: yeah just got lazy

rustaceanrob:
I added Tor to the `memory.rs` example. Could maybe use a rename

oleonardolima:
> I added Tor to the `memory.rs` example. Could maybe use a rename

Yep, I mean was it being used for other purposes, or it's fine to "steal" it for Tor purposes ?

rustaceanrob:
Yeah I think it can demonstrate both. The in-memory peer database should be rare or never used.

PR #77, opened 2024-08-15
fix(db): add schema versioning to sql

rustaceanrob:


oleonardolima:
partially unrelated question: Is it the standard to have the queries/statements in the code instead of a `.sql` file?

I don't have much experience with sqlite+rustqlite, that's why the question 😂

rustaceanrob:
Potentially that could be better for these schema queries, but it's nice to have the queries directly in rust when you use variables

PR #183, opened 2024-10-30
lib: update `bip324` dependency

rustaceanrob:


Cargo.toml
```diff
@@ -19,7 +19,10 @@ bitcoin = { version = "0.32.0", features = [
     "serde",
     "rand-std",
 ], default-features = false }
-bip324 = { version = "0.3.1" }
+bip324 = { version = "0.4.0", default-features = false, features = [
+    "std",
```
nyonson, line 23: technically `std` includes `alloc` right? But I guess this doesn't hurt?

nyonson (approved):

Cargo.toml
rustaceanrob: Wasn't entirely sure and was playing safe instead of sorry lol

PR #249, opened 2025-01-03
Fetch a range of headers from client

rustaceanrob:
Especially for consumers that have strict chain ordering requirements, like LDK, batching requests for headers should greatly speed up sync times.

### DIFF
--- a/src/core/client.rs
+++ b/src/core/client.rs
@@ -74,6 +74,10 @@ impl EventSender {
     /// # Errors
     ///
     /// If the node has already stopped running.
+    ///
+    /// # Panics
+    ///
+    /// When called within an asynchronus context (e.g `tokio::main`).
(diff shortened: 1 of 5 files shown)


src/core/client.rs
```diff
@@ -192,6 +209,25 @@ impl EventSender {
...
     }
 
+    /// Get a range of headers by the specified range.
+    ///
+    /// # Errors
+    ///
+    /// If the node has stopped running.
+    pub async fn get_header_range(
+        &self,
+        range: Range<u32>,
```
rustaceanrob, line 219: Need to try the more general `impl RangeBounds`

rustaceanrob: `RangeBounds` is not object safe, so it may not be used with `Box`

PR #266, opened 2025-01-14
Fix readme example and update MSRV pinning

nyonson:


rustaceanrob:
ACK e26ac29c02b24d81919dd39c518d03846b5adf78

PR #268, opened 2025-01-15
Separate `Warning` into its own channel, refactor transaction rejection naming

rustaceanrob:
Previous to this PR, the relationship between `Log` and `Warning` is that a `Warning` is contained in a `Log`. Because the log channel had limited capacity, a `Warning` event can technically be missed if the receiver is lagging behind. To compensate for this, transaction rejections at the network level were put into the `Event` enum, which have an unbounded channel to miss any `Event`. In reality though, a user would probably not want to miss any `Warning` at all, including transaction rejections. 

As an aside, the term "failure" in `FailurePayload` is not correct, as a single node may reject a transaction while others accept it. Instead, the `FailurePayload` is renamed to `RejectPayload` to more accurately reflect the situation.

This PR introduces the `RejectPayload` into the transaction rejection variant on `Warning`. In addition, the `Warning` variant is removed from `Log`, and `Warning` events are given their own channel. Now there is a clear separation of concerns between a `Log`, `Warning`, and `Event` without ambiguity of what data should be in each type.

@nyonson the bulk of the refactor are one line changes to account for this new warning channel sender/receiver. I tried to break the commits into easily reviewable parts, minus the head commit that has to do some heavy lifting. Of note in that one, any time the node needs to talk to the client, it must do so through the `Dialog`  struct. To redirect the `Warning` messages to the new channel instead of with `Log`, all that changes is the implementation of `Dialog::send_warning`.

### DIFF
--- a/src/core/dialog.rs
+++ b/src/core/dialog.rs
@@ -5,12 +5,21 @@ use super::messages::{Event, Log, Progress, Warning};
 #[derive(Debug, Clone)]
 pub(crate) struct Dialog {
     log_tx: Sender<Log>,
+    warn_tx: UnboundedSender<Warning>,
     event_tx: UnboundedSender<Event>,
 }
 
 impl Dialog {
-    pub(crate) fn new(log_tx: Sender<Log>, event_tx: UnboundedSender<Event>) -> Self {
-        Self { log_tx, event_tx }
+    pub(crate) fn new(
+        log_tx: Sender<Log>,
+        warn_tx: UnboundedSender<Warning>,
+        event_tx: UnboundedSender<Event>,
+    ) -> Self {
+        Self {
+            log_tx,
+            warn_tx,
+            event_tx,
+        }
     }
 
     pub(crate) async fn send_dialog(&self, dialog: impl Into<String>) {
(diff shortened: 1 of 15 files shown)


src/core/dialog.rs
```diff
@@ -40,7 +49,7 @@ impl Dialog {
     }
 
     pub(crate) async fn send_warning(&self, warning: Warning) {
```
rustaceanrob, line 51: Need to remove this `async`

src/core/dialog.rs
```diff
@@ -48,7 +48,7 @@ impl Dialog {
         let _ = self.log_tx.send(Log::Dialog(message)).await;
     }
 
-    pub(crate) async fn send_warning(&self, warning: Warning) {
```
nyonson, line 42: I believe the `send` call is async right? Kinda looks weird dropping the wrapper async if that is the case. Are there any issues in the tokio runtime with this though?

rustaceanrob: [`UnboundedSender<T>`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedSender.html) is sync. CI would hopefully fail for not `await`ing a future

nyonson (approved):
ACK 94dde9049ca08891974d5147399cfaac67ff789f

Thanks for the write up and commit break down, super helpful. I need to dig into these channels more, but makes sense to me from a high level.

PR #270, opened 2025-01-15
Remove async from `send_event`

rustaceanrob:
`Dialog::send_event` also makes use of the unbounded mpsc channels. From the docs:
> This method is not marked async because sending a message to an unbounded channel never requires any form of waiting.

There is no need to wait for more room on the stack or heap because the unbounded channels don't have a designated memory space. The bounded channels have to wait for sending an item because there may or may not be space in the channel. Conceptually a "channel" is just a queue in memory with one type that may place things there and another that can take things. 

Follows up to the previous `send_warning` refactor @nyonson


nyonson (approved):
ACK 7ea99c11fa574e45321f3e506c64320bcd89eea3

PR #271, opened 2025-01-15
Improve logging in integration tests

rustaceanrob:
First commit I remove the `async` from the node building functions. A very long time ago some of the start-up code was async, but I refactored it such that the builder is synchronous.

The logging in the integration tests hasn't gotten a lot of love by me in a while, but this utility function prints all the available info as the nodes sync, so will be useful for debugging in the future. 

cc @nyonson


nyonson (approved):
Ack 37a1ba24a21e767c37884c98d8f72acefbdd7cea

Nice, looks super helpful.

PR #272, opened 2025-01-15
Drop cargo configs

nyonson:
Instead of relying on dirty-ing a checked in file (`.cargo/config.toml`), switch to a feature flag based flow to enable the tokio debug console. This avoids accidental thrashing in the checked in file. It also allows a discoverable path from the justfile for new developers.

This is my first foray into the `build.rs` system, but it seems pretty straightforward. Also trying to be better about conventional commit messages.

Edit: Dropped all changes except for removing the cargo configs for now.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,0 @@
-# uncomment to use tokio-console
-
-# [build]
-# rustflags = ["--cfg", "tokio_unstable"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,9 +58,7 @@ tracing-subscriber = "0.3"
 tempfile = "3"
 tokio = { version = "1.37", default-features = false, features = [
     "full",
-] } # add feature "tracing" to use the console
-# Enable the tokio-console task and poll observations
-# console-subscriber = "0.3.0"
+] }
 
 [lib]
 name = "kyoto"


rustaceanrob:
I'm all in favor of simplifying the configurations in the repo. Having a look at this change there's a lot I don't quite understand. For one, since crate features are public I'm not a fan of having a feature that would only be useful for developers. If folks want to profile, they can bring what they need into their dependencies on their own. I think this is enough motivation to instead maintain a `dev` branch, and have it build the profiler by default. That way only developers actively working on the repository would have access to the console. 

For this PR, can we 1. remove the  `.cargo` folder 2. remove the comments on L61-63 in `Cargo.toml` 3. revert the other stuff like remove the `just` and the build script.

Thanks for drawing attention to this, should definitively not be in master.

nyonson:
Good point on the public feature flags, didn't think of that. I'll dial this back to just dropping the config for now.

I am curious about a system that uses build.rs and maybe environment variable to enable certain developer features like the debug console. Then a just recipe could set them for the dev and it would have a nice discoverable interface. But feels a little much at the moment.

rustaceanrob:
There's also a couple comments in `Cargo.toml` on L61-63 that don't really apply now that we are scraping the `.cargo` configurations

nyonson:
> There's also a couple comments in `Cargo.toml` on L61-63 that don't really apply now that we are scraping the `.cargo` configurations

Oops, missed those. Cleaned up in recent commit.

rustaceanrob:
ACK 698bf42a7fa787e8102fd502fdb000182ef0eaaf

Thanks

PR #273, opened 2025-01-16
Remove invalid check for MTP during fork

rustaceanrob:
The `last_median_time_past_window` function assumes we are trying to extend the current chain with the new headers we receive. Before this patch, in `sanity_check` the MTP was evaluated even in a fork scenario, which is not a valid MTP check.

@nyonson blending a refactor with a functional change here but removing the variable reassignment is not too bad to track in the diff. Trying to make the iterator easier to reason about.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -372,9 +372,11 @@ impl<H: HeaderStore> Chain<H> {
 
         // The headers have times that are greater than the median of the previous 11 blocks
         let mut last_relevant_mtp = self.header_chain.last_median_time_past_window();
-        if !header_batch
-            .valid_median_time_past(&mut last_relevant_mtp)
-            .await
+        // If this batch does not extend from the current tip, the headers provided by `last_median_time_past_window` are invalid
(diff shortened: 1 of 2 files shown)


src/chain/chain.rs
```diff
@@ -372,9 +372,10 @@ impl<H: HeaderStore> Chain<H> {
 
         // The headers have times that are greater than the median of the previous 11 blocks
         let mut last_relevant_mtp = self.header_chain.last_median_time_past_window();
-        if !header_batch
-            .valid_median_time_past(&mut last_relevant_mtp)
-            .await
+        if self.tip().eq(&header_batch.first().prev_blockhash)
```
nyonson, line 375: I think a comment for this equality check would be helpful since wasn't immediately obvious to me what it was doing.

rustaceanrob: Added the briefest comment I could come up with

nyonson (approved):
ACK 4883b341382a4200a0be62c1e77806d63db156a1

PR #274, opened 2025-01-16
Implement `Median` for slice types with macro

rustaceanrob:
Attempting to remove the repeated logic here, but also trying to avoid the trait hell that is required for implementing `Median` in a generic way. I extracted the logic into a macro which is reusable for any type of interest to implement `Median` 

cc @nyonson

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -24,50 +24,29 @@ pub trait Median<T> {
     fn median(&mut self) -> T;
 }
 
-impl Median<i64> for Vec<i64> {
-    fn median(&mut self) -> i64 {
-        self.sort();
-        let len = self.len();
-        if len == 0 {
-            0
-        } else if len % 2 == 1 {
-            self[len / 2]
-        } else {
-            let mid = len / 2;
-            (self[mid - 1] + self[mid]) / 2
+macro_rules! impl_median_for {
+    ($t:ident) => {
+        impl Median<$t> for [$t] {
+            fn median(&mut self) -> $t {
+                self.sort();
+                let len = self.len();
+                if len == 0 {
+                    0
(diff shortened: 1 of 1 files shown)


nyonson:
Dang is there really no good trait combo to describe "numbers"? I am a macro noob so I am gonna just play around with this one to learn it.

rustaceanrob:
I would say a number is described by the following `T: Copy + PartialOrd + Default + std::ops::Add<Output = T> + std::ops::Div<Output = T> + From<u8>`, which is horrible. I prefer the macro because the arithmetic is clear and there are no weird `From::(2)` type shit going on

nyonson:
> I would say a number is described by the following `T: Copy + PartialOrd + Default + std::ops::Add<Output = T> + std::ops::Div<Output = T> + From<u8>`, which is horrible. I prefer the macro because the arithmetic is clear and there are no weird `From::(2)` type shit going on
> 
> ref: https://internals.rust-lang.org/t/please-can-we-add-a-basic-num-trait/16449

Wow yea. I guess we could shove that in an alias and boil it down to just the bare minimum requirement for the function (which could be what you have already), but we till wouldn't get away from the weir `From::(2)` stuff.

nyonson:
Yea, I still find the macro syntax wierd, but that is me just not knowing. I think the macro is essentially doing what the compiler would too for the trait bound?

nyonson (approved):
Concept ACK 73beeaa43d0541165075f3865056527cb6ec35e0

I am gonna take some time to finally learn macros

rustaceanrob:
>  think the macro is essentially doing what the compiler would too for the trait bound?
Yeah, the compiler will re-implement the trait for each type that ends up using it

PR #275, opened 2025-01-17
Update stale block time after resetting connections

rustaceanrob:
If a block is considered stale, the `advance_state` function will reset connections with all peers. To reset the "staleness" of a block, the `LastBlockMonitor::update` function must be called. Currently that is done upon receiving new block headers, but no headers will ever be sent, as `advance_state` disconnects the peers before they can send them. That causes a loop that makes no progress. Here we patch this to reset the "staleness" of a block after disconnecting from all peers.

For context, the `LastBlockMonitor` is used to prevent a block withholding attack, where all connected peers refuse to advertise new blocks to us. After 30 minutes if there is no block found, we open new connections to see if there was a block being withheld. @nyonson

### DIFF
--- a/src/core/node.rs
+++ b/src/core/node.rs
@@ -188,7 +188,7 @@ impl<H: HeaderStore, P: PeerStore> Node<H, P> {
         let mut client_recv = self.client_recv.lock().await;
         loop {
             // Try to advance the state of the node
-            self.advance_state(&last_block).await;
+            self.advance_state(&mut last_block).await;
             // Connect to more peers if we need them and remove old connections
             self.dispatch().await?;
             // If there are blocks we need in the queue, we should request them of a random peer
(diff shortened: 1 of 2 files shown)


src/core/node.rs
```diff
@@ -476,6 +476,7 @@ impl<H: HeaderStore, P: PeerStore> Node<H, P> {
                         .send_dialog("Disconnecting from remote nodes to find new connections")
                         .await;
                     self.broadcast(MainThreadMessage::Disconnect).await;
+                    last_block.update();
```
nyonson, line 479: Maybe worth dropping a comment here why this is being updated? At first blush definitely non-intuitive to see and "update" when you are in a "stale" state code block. In this scenario it is more like a "reset" to start over right?

rustaceanrob: I want to try to avoid too many "what" comments throughout the repo since there is so much code as it is. You are right that the method should speak for itself, and it's more of a `reset`. I will do that refactor and omit the comment.

rustaceanrob:
This patch resolved an instance of a stale block on `Testnet4` and has been running with no issues since. Going to go ahead and merge

PR #278, opened 2025-01-23
Refactor the `Warning` type

### LINKED ISSUE #277: `Warning` type could use some work
`Warning` hasn't gotten a lot of love in descriptiveness. Partially this was due to consideration for language bindings, but it is no longer the case that `bdk-ffi` would use the type directly. And as a general principle, the Rust code should be as expressive as possible and the bindings can water it down as needed.
 
Some steps to improve the type
- [x] `NotEnoughConnections` could be renamed to the more concise `NeedConnections` and have fields `required`, `current`
- [x] `UnlinkableAnchor` has a strange name. It could be more direct, like `InvalidStartHeight`. Ideally it has fields `configured`, `lowest_stored` to reflect the height mismatch
- [x] Stylistically the `TransactionRejected` should have a named field `payload` to add documentation for the field
- [x] `FailedPersistance` is spelled wrong
- [ ] Maybe the `PotentialStaleTip` could include the height it is stuck on, although not sure what a user could do with that

rustaceanrob:
Closes #277 

Cleaning up the `Warning` type here to resolve a typo, refactor to clearer names, and add additional context to warnings so the developer might act on them. Also adding some additional context to a `HeaderPersistenceError`.

Each refactor is broken down commit by commit so it should be okay to review. Only context needed is in the `CannotLocateHistory` variant, which throws the user configures a checkpoint that is at a height lower than what is stored in the SQL database on start up. This prevents the user from going backwards _if they have already synced before_, which is a pretty soft requirement. For the first sync one can start from anywhere.


rustaceanrob:
Dude fuck the MSRV. I can't use `first_key_value`?? Lol

rustaceanrob:
Thought about it for a second. While the last commit would be useful, the `load_headers` function needs work anyway. Going to remove the last commit from this PR, so the whole context about `CannotLocateHistory` is no longer relevant for this PR.

Cargo.lock
```diff
@@ -1641,7 +1641,7 @@ dependencies = [
 
 [[package]]
 name = "kyoto-cbf"
-version = "0.7.0"
+version = "0.8.0"
```
nyonson, line 1644: Is this some leftovers?

nyonson (approved):
ACK bb934c3d7ee4c179905e0cef4b61bea6d53322a1

PR #282, opened 2025-02-03
Remove the MTP check

### LINKED ISSUE #265: Testnet4 block header check fails
The median time passed check is failing on Testnet4. Requires looking at the documentation for Testnet4.

rustaceanrob:
The current MTP check is hard to reason about and caused syncing issues over Testnet4. In a greater effort to refactor how `HEADERS` messages are evaluated, I am opting to remove the MTP check and add it back later if deemed worthwhile.

Closes #265 

cc @nyonson


nyonson (approved):
ACK 97a7a485fca72103d7f0695f95773caea44b5588

PR #283, opened 2025-02-04
Make CPU bound tasks synchronous in `Chain`

rustaceanrob:
From my original understanding of `async` runtimes I thought you could slap an `async` on anything you expect to take some time to complete. After further reading, I don't think this is the correct approach for CPU-bound work that does not benefit from concurrency (unless by use of multiple threads but that is besides the point here). The additional context switching of adding a future that may or may not poll as ready might even add some overhead. Here I remove the `async` from CPU-bound functions to remove any unnecessary context switching.

Happy to let this one marinate for a little if there is some interest in actually profiling if this does anything. Although I don't think there will be any noticeable effect, I think it is good practice to remove the `async` from these functions. 

ref:  https://users.rust-lang.org/t/should-you-only-use-futures-for-i-o-bound-tasks/100784

cc @nyonson

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -318,7 +318,7 @@ impl<H: HeaderStore> Chain<H> {
             return Ok(());
         }
         // We check first if the peer is sending us nonsense
-        self.sanity_check(&header_batch).await?;
+        self.sanity_check(&header_batch)?;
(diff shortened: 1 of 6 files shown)


nyonson (approved):
ACK 3603114c540ac5f9f0c0dad35c4c6ed7b01c3055

PR #284, opened 2025-02-05
Remove `dns` feature and enable DNS by default

rustaceanrob:
The original intention with the DNS feature was to enable the _ultra_ privacy-conscious users to circumvent DNS seeding, which would technically reveal in plaintext that the device is attempting to connect to bitcoin nodes. A feature in this case is overkill, as the binary size would be hardly effected if disabled, and the feature _should_ be used by the vast majority of users. I think a better approach in the future would be to disable DNS via the `NodeBuilder`. I am opting out of adding this configuration on `NodeBuilder` though, unless it is explicitly requested with some decent rationale.

Further still, if someone is actually doing traffic analysis, not all peers will connect over V2 transport at the moment anyway. 

Also to increase privacy in terms of the resolver handling our DNS queries, a custom resolver could be configured, which is something that should be easy enough to add to the `NodeBuilder`.

cc @nyonson

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,8 @@ arti-client = { version = "0.21.0", features = [
 tor-rtcompat = { version = "0.21.0", features = ["tokio"], optional = true }
 
 [features]
-default = ["database", "dns"]
+default = ["database"]
 database = ["rusqlite"]
-dns = []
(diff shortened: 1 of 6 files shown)


nyonson (approved):
ACK bf45020871bd46a1e9e9bdaa3392f522c992616b

PR #286, opened 2025-02-07
Introduce `PeerId` for stricter type meaning across the crate

rustaceanrob:
There is a `u32` nonce that is used to identify peers when we open a connection to them and want to send a specific peer a message. Refactoring this to a wrapper type is hopefully easier to read, especially for folks not familiar with the library but would potentially want to submit a patch. 

You'll notice I used `wrapping_add` to prevent an overflow as well. Funny enough Bitcoin Core had the same problem of using plain old `u32` math that could be exploited.

cc @nyonson

### DIFF
--- a/src/core/channel_messages.rs
+++ b/src/core/channel_messages.rs
@@ -11,6 +11,8 @@ use bitcoin::{
 
 use crate::core::messages::RejectPayload;
 
+use super::PeerId;
+
 #[derive(Debug, Clone)]
 pub(crate) enum MainThreadMessage {
     GetAddr,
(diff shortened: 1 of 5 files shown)


src/core/channel_messages.rs
```diff
@@ -11,6 +11,8 @@ use bitcoin::{
 
 use crate::core::messages::RejectPayload;
 
+use super::PeerId;
```
nyonson, line 14: I don't see `super` used much beyond unit test modules pulling in the code they are testing. I think it is a little less maintainable then a full crate namespace? Not a blocker or anything, just curious on the style.

nyonson (approved):
ACK cf0617c2cb3a025af0573b9d8147e7486376e710

src/core/channel_messages.rs
rustaceanrob: Just the LSP auto complete. Could potentially make an issue for a greater refactor to tackle then. `super` is used quite a bit across the crate.

PR #287, opened 2025-02-07
Configure DNS resolver

rustaceanrob:
If you think CloudFlare is spyware, you might want to change the DNS resolver from `1.1.1.1`

cc @nyonson

### DIFF
--- a/src/core/builder.rs
+++ b/src/core/builder.rs
@@ -1,3 +1,4 @@
+use std::net::{IpAddr, SocketAddr};
 use std::{path::PathBuf, time::Duration};
 
 use bitcoin::Network;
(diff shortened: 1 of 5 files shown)


nyonson (approved):
ACK 7f54f30441564ba2d833957d011fe91f3d04a61c

rustaceanrob:
Might have over-engineered my way out of the MSRV thing. Any type that is part of the node configuration must implement `Default`. It would've been nice to use the `const` constructor, but another approach is to make a new type that implements the `Default` socket address for the DNS resolver. Users don't need to know about this type and can just pass a `SocketAddr`, but internally I use this new wrapper type.

rustaceanrob:
Final tweak, 53 is required [by the spec](https://datatracker.ietf.org/doc/html/rfc1035#section-4.2), so it can just be implied when configured.

PR #288, opened 2025-02-08
Introduce `HeightMonitor` to track chain height relative to peers

rustaceanrob:
The previous design of determining if the chain is synced is a horrible artifact of moving from a proof-of-concept to an actual library. The `PeerMap` and `Chain` previously communicated via the `Node`, but there is no reason they can't share some common memory behind an `Arc` and `Mutex`.

Here I introduce a structure that keeps track of a peer and their height, and allows for the `Chain` to get the current maximum. I use the `tokio::sync::Mutex` to wrap this type, so the lock may be potentially held across `await` points if necessary. 

With this design, if a peer tries to grief the node by sending a ridiculous height, but cannot send the headers to reach that height, the node will disconnect from them and the height the malicious peer reported will no longer be used to determine if the node is synced.

cc @nyonson

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -17,6 +17,7 @@ use super::{
     checkpoints::{HeaderCheckpoint, HeaderCheckpoints},
     error::{BlockScanError, HeaderSyncError},
     header_chain::HeaderChain,
+    HeightMonitor,
 };
 #[cfg(feature = "filter-control")]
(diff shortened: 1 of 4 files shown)


rustaceanrob:
I have a machine running this patch on `Testnet4` and multiple reorgs have been handled properly + the node is in sync with the rest of the network. Going to self-ACK since this is a massive improvement in code quality with not much functional change imo.

PR #293, opened 2025-02-11
Move unit trait definitions into test module

### LINKED ISSUE #291: Unit should not implement backend traits
For testing, the `()` type implements `HeaderStore` and `PeerStore`. Unit is usable by anyone developing a Rust binary, so it may give users the wrong idea that these are safe to use. The implementation of these traits for the unit type should be removed. In the future, an in-memory database could be used for testing.

- [ ] Use a tempfile and remove all instances of `()` as a database

rustaceanrob:
Users should not be able to pass the unit type as a `HeaderStore` or `PeerStore` implementation because this would almost certainly cause syncing problems. However these implementations might still be useful as mock types for tests. As a compromise, these traits are defined in the test context. Unfortunately the definitions do not carry into the `tests` folder, only unit tests. I think this is fine though, as the integration tests should test the actual use of the library.

Closes #291 

cc @nyonson

### DIFF
--- a/src/db/error.rs
+++ b/src/db/error.rs
@@ -154,18 +154,3 @@ impl From<rusqlite::Error> for SqlHeaderStoreError {
         Self::SQL(value)
     }
 }
-
-/// Errors for the [`PeerStore`](crate) of unit type.
-#[derive(Debug)]
-pub enum UnitPeerStoreError {
-    /// There were no peers found.
-    NoPeers,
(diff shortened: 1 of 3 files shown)


nyonson (approved):
ACK 83a62497f6cbd90dc72d1bb262831609ffc2d080

Separation of public vs. private tests makes sense to me.

PR #294, opened 2025-02-11
Clean up the `HeaderStore` trait implementation

rustaceanrob:
The `HeaderStore` trait implementation is contrived to put it mildly. There is no reason to have a `write` and `write_over`, other than my initial laziness when building the project. All the trait should know is there is a mapping between `u32` and `Header`, and the implementation should handle conflicts. In the SQL context, this means replacing headers at a height if there is indeed a conflict.

Sadly this needs to stay a draft until an MSRV decision is made for the next release. You'll noticed the `clone()` of the header chain in the first commit which isn't good. I would instead add something more like `take` to grab the headers, but in that method I would likely want to use `last_key_value` (1.66) on `BTreeMap` , seen in the second commit

cc @nyonson

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -233,22 +233,12 @@ impl<H: HeaderStore> Chain<H> {
             .db
             .lock()
             .await
-            .write(self.header_chain.headers())
-            .await
-        {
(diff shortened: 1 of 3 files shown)


rustaceanrob:
Closing in favor of a better chain design in the future

PR #295, opened 2025-02-13
Improve `Filter` API

rustaceanrob:
There is a bit of trait hell going on within the `bip158` module. Something in the `BlockFilter` type is generic over `Read`, but the type itself concretely uses `Vec`. The `Read` implementation of  `Vec` doesn't do any I/O so the result is infallible. We can just add an `expect` instead of propagating this error everywhere.

ref: [Use of `Vec`](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/bip158.rs#L116) followed by [the trait hell](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/bip158.rs#L280)

cc @nyonson

### DIFF
--- a/example/managed.rs
+++ b/example/managed.rs
@@ -74,7 +74,7 @@ async fn main() {
                         Event::IndexedFilter(mut filter) => {
                             let height = filter.height();
                             tracing::info!("Checking filter: {height}");
-                            if filter.contains_any(&addresses) {
+                            if filter.contains_any(addresses.iter()) {
                                 let hash = *filter.block_hash();
                                 tracing::info!("Found script at {}!", hash);
                                 let indexed_block = requester.get_block(hash).await.unwrap();
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -727,7 +727,7 @@ impl<H: HeaderStore> Chain<H> {
         #[cfg(not(feature = "filter-control"))]
         if !self.block_queue.contains(&filter_message.block_hash)
             && filter
-                .contains_any(&self.scripts)
+                .contains_any(self.scripts.iter())
                 .map_err(CFilterSyncError::Filter)?
         {
             // Add to the block queue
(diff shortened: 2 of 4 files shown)


rustaceanrob:
While I was at it, I changed the API signature of `match_any` to be more generic. All the collections implement `Iterator`, so limiting to `HashSet` is more of an internal implementation detail. I caught the `impl` bug in method signatures over the past couple months.

nyonson:
> There is a bit of trait hell going on within the `bip158` module. Something in the `BlockFilter` type is generic over `Read`, but the type itself concretely uses `Vec`. The `Read` implementation of `Vec` doesn't do any I/O so the result is infallible. We can just add an `expect` instead of propagating this error everywhere.
> 
> ref: [Use of `Vec`](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/bip158.rs#L280) followed by [the trait hell](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/bip158.rs#L280)
> 
> cc @nyonson

Hm, the links you provided are the same? Not sure if intended cause I don't see the Read trait usage. But in any case, is the idea that there will definitely not be I/O errors since no I/O is actually happening?

rustaceanrob:
Oops, [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/bip158.rs#L116) is the line showing the `Vec`. Yeah, for this particular use case, the in-memory readers will be infallible since the size of the data is known and there's no I/O

nyonson:
Feels a little bad ignoring the API, but yea, makes sense.

nyonson (approved):
ACK 081b38cbaff2cc10d1fc1041949c08d7ce74aaf5

rustaceanrob:
self-ACK 87a9403426d68e00e4b8add7f450bfa8d000dcb9

PR #299, opened 2025-02-15
Actually broadcast transactions

rustaceanrob:
To prevent spam, the peer to peer protocol actually requires a transaction be announced via an `inv` of the transaction hash, and if the remote node does not have that transaction, they respond with a `getdata`. Finally, the local sends a `tx` message with the full encoding. 

I am embarrassed to admit I had not researched this enough, and currently I don't believe previous versions would successfully broadcast transactions. I did test, however, this patch works on Signet by using a BDK wallet to generate the transaction, then checking mempool.space. A follow up test in the integration tests could be nice, but is non-trivial to implement.

Because this library doesn't have any users yet, and future wallets will almost certainly use SegWit for the discount, I decided to reject peers that have a protocol version lower than SegWit. This should be no issue as SegWit is adopted by virtually every client.

ref: [BIP 339](https://github.com/bitcoin/bips/blob/master/bip-0339.mediawiki) WTx inventory

### DIFF
--- a/src/core/channel_messages.rs
+++ b/src/core/channel_messages.rs
@@ -6,7 +6,7 @@ use bitcoin::{
         message_network::VersionMessage,
         ServiceFlags,
     },
-    Block, BlockHash, FeeRate, Transaction,
+    Block, BlockHash, FeeRate, Transaction, Wtxid,
 };
 
 use crate::core::messages::RejectPayload;
@@ -17,6 +17,7 @@ use super::PeerId;
 pub(crate) enum MainThreadMessage {
     GetAddr,
     GetAddrV2,
+    WtxidRelay,
     GetHeaders(GetHeaderConfig),
     GetFilterHeaders(GetCFHeaders),
     GetFilters(GetCFilters),
@@ -58,6 +59,7 @@ pub(crate) enum PeerMessage {
     #[allow(dead_code)]
     Pong(u64),
     FeeFilter(FeeRate),
+    TxRequests(Vec<Wtxid>),
 }
 
 #[derive(Debug, Clone)]
--- a/src/core/node.rs
+++ b/src/core/node.rs
@@ -42,7 +42,7 @@ use super::{
     FilterSyncPolicy, LastBlockMonitor, PeerId, PeerTimeoutConfig,
 };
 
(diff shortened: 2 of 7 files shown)


rustaceanrob:
Did more digging on this. It turns out that there is no code that explicitly rejects a transaction if it was sent before an `inv`. Having a look at the net processing logic, [it looks like the transaction will be evaluated](https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L4211C21-L4211C31) even if no `inv` was sent for that transaction. I confirmed with [this transaction](https://mempool.space/signet/tx/7065485260e255cb2e74abac9848e56a4b9019d02e7f7f19d8d4682819fe4356) that simply sending a `tx` message with no `inv` still results in confirmation. Will have to YOLO a little money on mainnet today to verify there as well.

rustaceanrob:
Confirmed with multiple transactions with multiple peers on mainnet that a transaction still propagates if sent directly by `tx`. This patch is more or less cosmetic.

nyonson:
I'll sink into this just cause I wanna learn more about the details of the journey here, but may be nice to update the commit message to be be a little more in line with the findings?

rustaceanrob:
Rebased to a single commit because both commits were required for the fix and aren't independent. Also updated the commit message to describe more mechanically what's going on. 

The following function in the `inv` processing:

```cpp
GenTxid ToGenTxid(const CInv& inv)
{
    assert(inv.IsGenTxMsg());
    return inv.IsMsgWtx() ? GenTxid::Wtxid(inv.hash) : GenTxid::Txid(inv.hash);
}
```

Convinced me to approach the PR a bit differently. I think it is fair to reject any client that is not at least at the SegWit version. I think coming across a client that is not SegWit _and_ supports CBF is improbable if not impossible. Assuming we are connected to a SegWit client, as long as we send the `WTx` message with a `wtxid`, this bit of code proves that the inventory hash will be processed as a WTXID. Since this is the case, I updated the `GetData` logic to only look for `WTx` requests.

rustaceanrob:
Confirmed this worked using commit 8f41b89f22fb3f5a740f0c838e1c68e6948e346a

PR #300, opened 2025-02-16
Allow for compiler optimizations and logging configuration

### LINKED ISSUE #298: Logging requires heap allocations
Reading the discussion on [LDK node](https://github.com/lightningdevkit/ldk-node/pull/407), the `format!` macro allocates on the heap to create the strings. The `Log::Dialog` message is mostly useful for developers, but could be optimized away for a production application. This could either go behind a feature or a configuration that can be optimized way by the compiler.

rustaceanrob:
Closes #298 

The `format!` macro requires a heap allocation, and the `&str` references are converted to `String` when sent as a `Log::Dialog` message. Because some applications might be more memory hungry than others, there should be a way to stop the node from allocating these `String` entirely. I thought about having this as a crate feature, but the language bindings used by BDK would not be able to make use of this performance gain, since they ship a binary. I think I strike the best of both worlds here and use a runtime variable that can still be optimized away by the compiler.

To prove this works, I made a simple example. Check out this in the [Rust playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2021) and select "ASM" using the three dots on the top left.

```rust
macro_rules! conditional_run {
    ($foo:expr, $expr:expr) => {
        match $foo.condition {
            Condition::Run => { $foo.do_foo($expr); }
            Condition::Skip => {} 
        }
    };
}

enum Condition {
    Run,
    Skip,
}

struct Foo {
    condition: Condition
}

impl Foo {
    fn do_foo(&self, message: String) {
        println!("{message}");
    }
}

fn main() {
    let condition = Condition::Skip;
    let foo = Foo { condition };
    conditional_run!(foo, format!("Allocated String!")); 
    println!("No allocation occurred!");
}
```

The compiler can remove the branching and either completely ignore the expression, thus not allocating the `String`, or execute the expression and send a debug message. Only downside to this approach is the `Log::Dialog` is still a member of `Log` even though we don't need it. Not sure how to get rid of it without introducing compile-time options though.

### DIFF
--- a/example/managed.rs
+++ b/example/managed.rs
@@ -2,9 +2,9 @@
 //! many possible scripts to check. Enable the `filter-control` feature to check filters
 //! manually in your program.
 
-use kyoto::core::messages::{Event, Log};
+use kyoto::core::messages::Event;
 use kyoto::{chain::checkpoints::HeaderCheckpoint, core::builder::NodeBuilder, Client};
-use kyoto::{AddrV2, Address, BlockHash, Network, ServiceFlags, TrustedPeer};
+use kyoto::{AddrV2, Address, BlockHash, LogLevel, Network, ServiceFlags, TrustedPeer};
 use std::collections::HashSet;
 use std::{net::Ipv4Addr, str::FromStr};
 
@@ -45,6 +45,8 @@ async fn main() {
         .anchor_checkpoint(checkpoint)
         // The number of connections we would like to maintain
         .num_required_peers(1)
+        // Omit informational messages
+        .log_level(LogLevel::Warning)
         // Create the node and client
         .build_node()
         .unwrap();
@@ -63,9 +65,7 @@ async fn main() {
         tokio::select! {
             log = log_rx.recv() => {
                 if let Some(log) = log {
-                    if let Log::Dialog(log) = log {
-                        tracing::info!("{log}")
-                    }
+                    tracing::info!("{log}");
                 }
             }
             event = event_rx.recv() => {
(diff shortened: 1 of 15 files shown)


rustaceanrob:
While I was at it I changed `Log::Dialog` to a more reflective name in `Log::Debug`. I think this combined with `LogLevel` will be clear which message type is being omitted.

nyonson:
OK just runnin it back to make sure I get it, you are adding a conditional here based on a flag so the compiler can be smart and just toss out branches that will never be executed. Without this, things were always getting allocated on the heap even if the client wasn't running debug mode?

I think my one big question is how does the compiler optmise a runtime flag? In your playground example (btw pretty dope that godbolt asm stuff is just built right in to rust playground) won't the compiler pick up on the `condition` cause it is in face available at compile time (aka hardcoded)?

rustaceanrob:
Right, the compiler is smart enough to remove branches that will not execute. Even though the log level is set at runtime, it is _known_ for the entire duration of `Node::run`, so the compiler can still remove unused branches. To prove this to yourself you can checkout the branch and slap `dbg!` around any of the `format!` statements and see that it gets executed with `LogLevel::Debug` and does not execute with `LogLevel::Warning`.

rustaceanrob:
Also just realized I normally compile all examples in release, so might need to slap a `--release` for the compiler to do it's magic.

nyonson (approved):
ACK 8af5be1abf5259b5edf2c763d94041cc0462572d

Was a little thrown by the optimizations, but makes sense that a consistent conditional guard will protect against unnecessary log allocations.

PR #301, opened 2025-02-18
Use `Arc` to clone `Dialog`

rustaceanrob:
I think the proper "rustonic" thing to do here is use a reference counted pointer to anything that needs a `Dialog` to communicate with the client. I change all instances of `Dialog` with `Arc<Dialog>` and explicitly clone using `Arc::clone` to indicate the clone is cheap. Also verified that the recent log changes still omit unnecessary allocations.


rustaceanrob:
Self ACK as I think this one is pretty uncontroversial

PR #302, opened 2025-02-18
Use declarative naming for methods in builder

rustaceanrob:
This is mostly an LSP related thing, where I think removing the verb for most methods makes them easier to find. Trying to push in anything breaking before [this is merged](https://github.com/bitcoindevkit/bdk-ffi/pull/591)

cc @nyonson


rustaceanrob:
Another mostly cosmetic change. Better to break the API now than later.

PR #308, opened 2025-03-13
build: new msrv job to isolate build tools

nyonson:
The msrv job checks just that the library code can be built with the MSRV's compiler, and that is it. All other build tooling functionallity doesn't need to be gated on the MSRV since it doesn't effect consumers. We specifically would rather use cargo's V3 resolver by default to generate our lockfiles.


rustaceanrob:
Makes sense

ACK 6a341bc9ab0c956ab4b5e42151ccb1f8a7a663bf

PR #309, opened 2025-03-13
chore: pull readme badge from kyoto repo, not bdk

nyonson:
Perhaps we want to change the name of the "Build and Test" job to just "CI" like BDK has it (I like the simplicity). But in either case, don't need *their* CI status on this readme.


rustaceanrob:
Caught red handed

ACK 3f18114829744e96d7dbb11651ce32909fb07de9

Issue #310, opened 2025-03-14
Break up the core module


nyonson:
This will mostly be an exercise in shifting things around, naming, and perhaps a tweak or two to the public API after review. But the goal is to have clearer modules and more testable interfaces.

The current state of the `core` module.

```
+--------------------+---------------------+-------------------------+
| File               | Primary Types       | Primary Functions       |
+--------------------+---------------------+-------------------------+
| broadcaster.rs     | Broadcaster         | queue(), add()          |
| builder.rs         | NodeBuilder         | build(), add_scripts()  |
| channel_messages.rs| MainThreadMessage   | message definitions     |
|                    | PeerThreadMessage   |                         |
| client.rs          | Client, Requester   | broadcast_tx(), get_*() |
| config.rs          | NodeConfig          | configuration           |
| dialog.rs          | Dialog              | send_info(), warn()     |
| error.rs           | NodeError, etc.     | error handling          |
| messages.rs        | Event, Log, Warning | message definitions     |
| mod.rs             | -                   | re-exports              |
| node.rs            | Node, NodeState     | run(), handle_*()       |
| peer_map.rs        | PeerMap             | connect(), live()       |
+--------------------+---------------------+-------------------------+
```

I want to explore breaking up `core` into a few new modules and perhaps promoting some of the most shared logic to top level files under `src/`.

(I'll admit I have a personal bias against modules called `core`, `common`, `util` etc. since they generally trend towards large dumping grounds of hard to test code.)

PR #311, opened 2025-03-14
chore: deprecate core module

nyonson:
Remove the catch-all core module by promoting the implementation to the top level kyoto module. Some refinements were made by pushing things down into existing modules, like the peer_map to network. I still some smaller questions that can be dealt with later, like does broadcaster need to be its own module? Should that also live in network?

Related to #310


doc/DETAILS.md
```diff
@@ -81,18 +81,23 @@ The wallet required 12 block downloads, and took 5 minutes.
 
 This section details what behavior to expect when using Kyoto, and why such decisions were made.
 
-### Peer Selection
```
nyonson, line 84: Is jumping from H1 to H3 a style choice?

example/testnet4.rs
```diff
@@ -1,6 +1,6 @@
 //! Usual sync on Testnet.
 
-use kyoto::{chain::checkpoints::HeaderCheckpoint, core::builder::NodeBuilder};
+use kyoto::{builder::NodeBuilder, chain::checkpoints::HeaderCheckpoint};
```
rustaceanrob, line 3: Not shown in the diff, but `managed` and `tor` examples will likely also need an upgrade. They require the additional `features` flag when building.

src/network/peer.rs
```diff
@@ -41,6 +40,42 @@ const HANDSHAKE_TIMEOUT: u64 = 4;
 
 type MutexMessageGenerator = Mutex<Box<dyn MessageGenerator>>;
 
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
+pub(crate) struct PeerTimeoutConfig {
```
rustaceanrob, line 44: I prefer having large types like `Peer` having their own self-contained file. Not sure if that is abnormal, but I also think the case can be made that these types (`PeerTimeoutConfig`, `PeerId`) should exist at the top level `mod` in `network`. Unless the case can be made strongly against it, I prefer these go to `network/mod.rs`

src/node.rs
```diff
@@ -39,14 +46,46 @@ use super::{
...
+        if let Some(time) = self.last_block {
+            return Instant::now().duration_since(time) > Duration::from_secs(THIRTY_MINS);
+        }
+        false
+    }
+}
+
+/// Should the node immediately download filters or wait for a command
+#[derive(Debug, Default)]
+pub enum FilterSyncPolicy {
```
rustaceanrob, line 81: I believe this could go to `lib.rs`

example/testnet4.rs
nyonson: Ah missed that ok, I am gonna try and add that feature flag matrix to CI in a separate PR too.

src/network/peer.rs
nyonson: I think both this and the `FilterSyncPolicy` comment are my golang showing. My instinct is to keep the types together and then export them when necessary, instead of defaulting to a crate-wide spot. I am not sure I have a good feel yet for managing `lib.rs`, but I though it could be used just to control the public api, including re-exporting types from internal modules. Is there a pattern for instead doing a top down approach?

nyonson:
Ok everything iron'd out in 3da587482139a0941c5f291cd7e6685c244eab47 except tor example, but that is broke on master right now too, will address separately.

src/lib.rs
```diff
@@ -123,11 +145,12 @@ pub use tokio::sync::mpsc::UnboundedReceiver;
...
-    crate::core::builder::NodeBuilder,
-    crate::core::client::{Client, Requester},
-    crate::core::error::{ClientError, NodeError},
-    crate::core::messages::{Event, Log, Progress, RejectPayload, SyncUpdate, Warning},
-    crate::core::node::{Node, NodeState},
+    crate::builder::NodeBuilder,
+    crate::client::{Client, Requester},
+    crate::error::{ClientError, NodeError},
+    crate::messages::{Event, Log, Progress, RejectPayload, SyncUpdate, Warning},
+    crate::network::PeerTimeoutConfig,
```
rustaceanrob, line 152: This was previously used internally and did not hit the public API, but I think another method on the builder to allow users to use the `PeerTimeoutConfig` type would be a good follow up

rustaceanrob:
We should probably remove the "Getting Started" section of the documentation in `lib.rs` since it now points to the rustlang `core` module. Or maybe point users towards `builder.rs`

rustaceanrob:
ACK 41a9ac6b90da62296f587a11bfa64dced44505e7

PR #314, opened 2025-03-28
Move `LastBlockMonitor` and `NodeState` out of `node` module

rustaceanrob:
The `node` module is getting cluttered and these are self-contained structures. Doing a bit of a stretch in encapsulation here, but I think since the time we see the last block is implicit to the node network connections, I would like it to be in `network`. The `NodeState` is currently very primitive, so I think `lib` is suitable. If `NodeState` starts to contain data or methods then it might belong in `node` again.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -146,7 +146,7 @@ pub use {
     crate::error::{ClientError, NodeError},
     crate::messages::{Event, Log, Progress, RejectPayload, SyncUpdate, Warning},
     crate::network::PeerTimeoutConfig,
-    crate::node::{Node, NodeState},
+    crate::node::Node,
 };
 
 #[doc(inline)]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -468,6 +468,39 @@ pub enum FilterSyncPolicy {
     Continue,
 }
 
+/// The state of the node with respect to connected peers.
+#[derive(Debug, Clone, Copy)]
+pub enum NodeState {
```
nyonson, line 473: This is a personal preference so not holding up this patch, but to me `NodeState` is so specific to node, I would favor keeping it there so that there is clear ownership. Node defines its states. I think this is inline with my general preference to use lib.rs as a "table of contents" which holds very little definitions itself.

nyonson (approved):
ACK 4a083257f9add9ea908330f6b3b8150c3fa2236c

src/lib.rs
rustaceanrob: I understand that reasoning definitely. It is used as part of the `Log` message though. If it does gain function in the future, I think it should go back into `node` for sure.

PR #315, opened 2025-03-28
build(dep): upgrade corepc node for zip dep

nyonson:
The zip dependency upgrades cleans up some broken MSRV-ness. Doesn't hurt to keep our dev deps up to speed too.


rustaceanrob:
ACK 6aca0e101ba909f29347a15929c0932096a26526

PR #317, opened 2025-03-28
build(dep): loosen tokio requirements

nyonson:
Since this is a library, let the consumer dictate the version of tokio which works for them. The library should operate against the v1 tokio interface to be as accommodating as possible.


rustaceanrob:
ACK 5d903fef76c4ea41b51bcabe0426833b291d5d6a

PR #320, opened 2025-03-31
Add signet, testnet4, mainnet checkpoints

rustaceanrob:
Adding recent heights and hashes for all networks. Verified these are correct locally, but could use another audit.


nyonson (approved):
ACK 55896de302283a68b3d11a2fdadec97e9b8cb128

Verified the mainnet ones on my node.

PR #325, opened 2025-04-05
Remove the error case for `match_any`

rustaceanrob:
As of Rust lang `1.0.0` the `&[u8]` type does not include any error cases in the `Read::read` function (instead there are panics). The bip158 module of `rust-bitcoin` involves a lot of error propagation, but ultimately the fallible function is `Read::read` on a slice. Removing the error case allows for a cleaner API both internally and externally for future silent payments users.

Read trait invocation: https://docs.rs/bitcoin/latest/src/bitcoin/bip158.rs.html#455
The buffer: https://docs.rs/bitcoin/latest/src/bitcoin/bip158.rs.html#470
Read implemented for `&[u8]`: https://doc.rust-lang.org/src/std/io/impls.rs.html#298-403

Note this is the same message in the commit so there is a reference.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -723,9 +723,7 @@ impl<H: HeaderStore> Chain<H> {
 
         #[cfg(not(feature = "filter-control"))]
         if !self.block_queue.contains(&filter_message.block_hash)
-            && filter
-                .contains_any(self.scripts.iter())
-                .map_err(CFilterSyncError::Filter)?
+            && filter.contains_any(self.scripts.iter())
         {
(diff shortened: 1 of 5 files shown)


rustaceanrob:
Shoot. I thought the call [here](https://docs.rs/bitcoin/latest/src/bitcoin/bip158.rs.html#455) was from the `Read` trait but it's actually a function on the struct that is being used. Within that there is a `read_exact` which is technically fallible. I still think this cannot error for `Vec<u8>` but perhaps if a filter is malformed and sent maliciously it can. Requires a little research.

PR #328, opened 2025-04-11
Introduce `Info` and separate info/debug strings

rustaceanrob:
#### Description

While it might be painful to have 4 different ways to handle messages, I think this design is worth the performance improvements if one would like to completely remove string allocations by omitting the debug messages. Informational messages can be useful but are far less chatty, and may potentially be expanded into the future (like positive matches for filters).

For a desktop application I don't see much harm in using the debug strings and writing to a log, but mobile users can't use the strings for much at all so they should have a way to omit them, but still get updates on the changes in state and such. 

#### Implementation
- Introduces a `info!` macro similar to `log!`, so informational messages aren't allocated with `LogLevel::Warning` set. 
- Adds a new receiver type and changes the `Log` enum to `Info`. The debug messages are emitted as strings and `Info` is an enum with the rest of the events

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,7 @@
 //! ```no_run
 //! use std::str::FromStr;
 //! use std::collections::HashSet;
-//! use kyoto::{NodeBuilder, Log, Event, Client, Address, Network, HeaderCheckpoint, BlockHash};
+//! use kyoto::{NodeBuilder, Event, Client, Address, Network, HeaderCheckpoint, BlockHash};
 //!
 //! #[tokio::main]
 //! async fn main() {
@@ -42,16 +42,13 @@
 //!     // Run the node and wait for the sync message;
 //!     tokio::task::spawn(async move { node.run().await });
 //!     // Split the client into components that send messages and listen to messages
-//!     let Client { requester, mut log_rx, warn_rx: _, mut event_rx } = client;
+//!     let Client { requester, mut log_rx, info_rx: _, warn_rx: _, mut event_rx } = client;
 //!     // Sync with the single script added
 //!     loop {
 //!         tokio::select! {
 //!             log = log_rx.recv() => {
 //!                 if let Some(log) = log {
-//!                     match log {
-//!                         Log::Debug(d) => tracing::info!("{d}"),
-//!                         _ => (),
-//!                     }
+//!                     tracing::info!("{log}");
 //!                 }
 //!             }
 //!             event = event_rx.recv() => {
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -445,9 +443,20 @@ macro_rules! log {
     ($dialog:expr, $expr:expr) => {
         match $dialog.log_level {
             crate::LogLevel::Debug => $dialog.send_dialog($expr).await,
-            crate::LogLevel::Warning => (),
+            _ => (),
+        }
+    };
+}
+
+macro_rules! info {
```
nyonson, line 451: Given the structure and relatively exhaustive things sent over the `info` channel, should it just be detached from the LogLevel setting entirely? I might be missing a benefit, but decoupling it would simplify this code (could drop this macro?) and some conditionals.

nyonson: Reviewing the 4 channels for the user interface at this point, I get where this change is coming from, but does feel a little pre-optimized. Does it make sense to combine the `Info` and `Event` channels to try and keep things simple?

rustaceanrob: There's an argument to be made for that, although I kind of envision the applications in 3 tiers:
1. desktop or terminal app that writes to a debug log
2. mobile or desktop application that adjusts UI components based on info/warnings
3. a sort of "headless" node that just runs in the background unless it dies, like in `ldk-node` there's really not much of a use in having info messages because it would be so deeply nested in the syncing process

Maybe it's over kill, but the `Progress` variant of `Info` in particular might be a little wasted if emitted for all types of apps. The nice thing about the `LogLevel` construct is you get exactly the memory/CPU cycles you sign up for, so I'd like to leave these coupled for now

rustaceanrob: >  Does it make sense to combine the Info and Event channels to try and keep things simple?

The original type of message was simply `NodeMessage`, and this caused havoc trying to match between things that were mostly unimportant and events like blocks. Doesn't feel so great to look at all these channels but the alternative is the inability to separate your program into "this part handles blocks" and "this part handles logs" type of thing

nyonson: Hm ok, I guess the "single receiver" constraint of channels kind of forces this opinionated separation of events.

nyonson (approved):
ACK 31b35ddf387214fc44ef359c62eb6d7279b6cf17

src/lib.rs
rustaceanrob: For a real-world use of constraining channels to specific messages, check out what we do it [`bdk_kyoto`](https://docs.rs/bdk_kyoto/latest/src/bdk_kyoto/lib.rs.html#89-98) to build updates for the wallet. If the `Event` included other things, it might be difficult to pass that one to the user. In that struct we really only care about blocks and reorgs anyway

nyonson: Yea, I'd stick to one channel you force the caller to make there own little channels in their side. If the split is obvious enough that each caller would create the same 3 internal channels, probably makes sense to define them on the Kyoto side

PR #330, opened 2025-04-12
Replace `CFHeaderChain` and `FilterChain`

rustaceanrob:
# Description

The filter headers, filter hashes, and if we checked the filters may now be closely coupled to the graph representation of the chain via `BlockTree`. This should ultimately save on memory and make syncing problems easier to reason about because the structure is the same for block header/filter header/filter messages. 

## Broken down by commit

**First**: The `CFHeaderChain` previously handled the state `CFHeader` messages. This state is re-implemented in the `FilterRequestState` struct that manages the pending requests, queued changes to the `FilterHeader` and `FilterHash`, as well as how many peers agree on a particular message. A `FilterCommitment` struct is also introduced to associate a `FilterHeader` with its corresponding `FilterHash` for each block. The `FilterCommitment` is added to the `BlockNode`, as well as a boolean if this filter has been checked or not. Finally a new iterator is added to iterate over all the `BlockNode` in the graph.

**Second**: Using the `FilterRequestState` and the newly defined methods, we can just use the `BlockTree` to handle if we checked a certain filter or not, and to generate the message for the next filters required. This removes the need for `FilterChain`, and is a low-lift

**Third**: Also using `FilterRequestState`, removed the `CFHeaderChain` and store all the filter commitments within the `BlockTree`. The checks for message validity are included in the body of the sync function, so they can be easily evaluated as early-return error conditions. Finally we can either queue these changes or merge them depending on how many peers have sent us the message for the requested range. Note that this becomes a simple equality check, as providing the blockhash of where a conflict originates wouldn't be useful. It is prohibitively costly to audit the filter for a block. It is best to disconnect from everyone and try again.

## Random caveats
- I unfortunately misread BIP 157 and thought the maximum headers to request was 100 when it's actually 1,000. That is fixed here.
- The `chain_update` calls got verbose. These calls are moved to a new function in `chain` which makes use of `log!` and `info!` to lazily evaluate the expressions. This removes the need for `Dialog::chain_update`
- One test had to be marked as `ignored`, but it was not properly implemented in the first place: When there is no previous filter header to compare to, the previous filter header in the `CFHeader` message is ignored. This occurs when starting a fresh sync. Once some `FilterHeader` are added to the chain, the previous filter header may be properly checked. This causes the failure, as before this check was being ignored, but now there is a mismatch in the expected and actual previous filter header. The `CFHeader` message is hardcoded in the test, and the value for the filter header came from Bitcoin Core. Core displays the filter header in the opposite endian-ness that the `hex::decode` is expecting, so the hardcoded value is wrong. The hardcoded value is also wrong because it is the filter header for a different block than expected. The values for this test should be re-generated, or the test can be removed in the future.


rustaceanrob:
Third commit had to be amended again cause I had a thought on the metro. Before there was not check that a peer was sending us the same commitments twice and artificially increasing the amount of peers that agree. Now the staged changes to the filter headers/hashes have an associated `PeerId` which is checked against in 400377e0db69270165f8bd0e7e8485b697309f47

I am leaning not to punish the peer in any way, it is possible to request data twice if a peer is slow

src/chain/chain.rs
nyonson, line 1: I am not sure of any benefits, or the cost, but was thinking how it might be nice to add a validation check to `check_send_block` that the block filter (and block filter header chain) match the given block data we just downloaded. Figure we have the data at that point (the hard part) so maybe worth doing a quick sanity check?

rustaceanrob: Can't do this the same reason we can't audit the block. Need the scriptPubKeys associated with the outpoints

nyonson: Ah, right forgot about that again, thanks.

src/chain/graph.rs
```diff
@@ -110,7 +110,8 @@ pub(crate) struct BlockNode {
     pub height: Height,
     pub header: Header,
     pub acc_work: Work,
-    pub filter_hash: Option<FilterHash>,
+    pub filter_commitment: Option<FilterCommitment>,
+    pub filter_checked: bool,
```
nyonson, line 114: `filter_checked` feels much more ephemeral than the rest of the data stored in BlockTree. I am wondering if it might make more sense to store it outside in its own layer, something a little closer to the user wallet. Haven't fully thought it through, and don't think it needs to change this patch, but curious if you thought about it already.

src/chain/chain.rs
```diff
@@ -47,8 +45,7 @@ const FILTER_BASIC: u8 = 0x00;
 #[derive(Debug)]
 pub(crate) struct Chain<H: HeaderStore> {
     pub(crate) header_chain: BlockTree,
-    cf_header_chain: CFHeaderChain,
-    filter_chain: FilterChain,
+    request_state: FilterRequestState,
```
nyonson, line 48: I have struggled to really digest everything going on `chain` at this point, bit of a beast. I don't think it has to be addressed in this patch since dropping a bunch of code and using BlockTree is definitely a step in the right direction. But leaving notes while its fresh on a few things I think are causing some confusion.`node`'s `next_stateful_message` is really capturing the state phases of `chain` and feels like it might be better as an enum in `chain` itself, or something along those lines. And in the same vein, I initially found it a confusing that `FilterRequestState` manages both the filter header sync as well as the filter sync. In my mind the filter header sync is just weird with its own little agreement logic, it should be isolated.

nyonson (approved):
ACK c3bb6c4c0eb91080954ff6992a41dcd1ee60c059

Left a few higher level thoughts, but don't wanna churn on this one too much and have faith in the tests.

src/chain/graph.rs
rustaceanrob: Although the `checked` flag is simple, I like how it leaves for iteration in a way that is coupled to the state of the header chain. i.e. maybe the filter is actually stored for blocks < 1..50 from the tip. It should also be noted that having this strict construction _should_ open up the door for parallel downloads of filters across multiple peers. However this is a massive BUT because consumers of the library expect blocks are in chronological order

PR #333, opened 2025-04-15
Filter on service flags when querying DNS

rustaceanrob:
I was researching how to run a DNS seeder that only filters on compact block filter nodes, and I was looking at both Sipa/Ava's implementations. I came across [this code comment](https://github.com/achow101/dnsseedrs/blob/main/src/dns.rs#L363) in Ava's implementation that made me scratch my head, but sure enough, you can prefix a DNS with the service flags you would like. So in your terminal:

`dig dig x49.seed.bitcoin.sipa.be` returns _only_ nodes that signal compact block filters. 

This is stupidly great for sync times because the initial seeding is only CBF peers. Have some follow-up ideas, but here is the main patch.


nyonson (approved):
ACK 2aface905f76f946a430e178bcdc46f51e5e9f21

PR #335, opened 2025-04-17
Use unbounded channels to avoid dropped requests

rustaceanrob:
The `UnboundedSender` offers two advantages over bounding the number of requests a client can make. First it drops the need to maintain an `async` and sync version of the client methods - `UnboundedSender::send` is synchronous. This is actually very convient for `ldk-node`, where the client will have to implement a synchronous trait.

Secondly there are some methods that are critical to succeed if the node is running. Adding a script must always work if the node is still online. Because it is unacceptable to miss some messages, there is basically two approaches: cache the failures and try again later, or use an unbounded channel. Since caching the failures will result in the exact same overhead as stored messages in a channel when the node is not ready to accept them yet, the unbounded channel is much easier to implement and reason about.

### DIFF
--- a/example/managed.rs
+++ b/example/managed.rs
@@ -96,6 +96,6 @@ async fn main() {
             }
         }
     }
-    let _ = requester.shutdown().await;
+    let _ = requester.shutdown();
     tracing::info!("Shutting down");
 }
--- a/example/rescan.rs
+++ b/example/rescan.rs
@@ -82,9 +82,9 @@ async fn main() {
             .unwrap()
             .require_network(NETWORK)
             .unwrap();
-    requester.add_script(new_script).await.unwrap();
+    requester.add_script(new_script).unwrap();
     // // Tell the node to look for these new scripts
-    requester.rescan().await.unwrap();
+    requester.rescan().unwrap();
     tracing::info!("Starting rescan");
     loop {
         tokio::select! {
@@ -107,6 +107,6 @@ async fn main() {
             }
         }
     }
-    let _ = requester.shutdown().await;
+    let _ = requester.shutdown();
     tracing::info!("Shutting down");
 }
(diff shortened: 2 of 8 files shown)


nyonson:
This interface looks good to me, so first point makes sense. But I'm not following the second. Does this change make anything more reliable? Seems like it just removes back pressure (which seems fine in this scenario).

rustaceanrob:
I incorrectly recalled this as a problem for the mpsc receivers but it is actually a `broadcast::Receiver` problem. Either way the message just fails on the send side, which results in this problem of caching failures. Instead of doing our own retry logic I think using the built-in is better here.

nyonson:
Aren't bounded and unbounded exposed to the same types of send failures? Sorry being slow here, just feels like I am missing something. I thought bounded are made `async` in case the buffer is full and it can send back pressure to the caller. And unbounded can still fail in some catastrophic way if memory runs out. But neither fail due to work done on the receiving side right?

nyonson (approved):
ACK 053261d423f40a4199b82c7cf504c38d3554e20b

rustaceanrob:
If a send to an unbounded channel fails then the channel is either closed or the receiver has been dropped. In both cases that would mean the node has stopped. In the bounded case you'll notice the `Sender::send` error variant has a `T` with it, which is intended to be used if the queue is currently full. In the bounded case the send can fail if the channel is closed/dropped, but can _also_ fail if the queue is full and the node is currently processing a request.

nyonson:
> In the bounded case the send can fail if the channel is closed/dropped, but can _also_ fail if the queue is full and the node is currently processing a request.

Ok, that mostly checks out with my understanding. But I don't think `Sender::send` fails if the queue is full, I think it just suspends?

PR #336, opened 2025-04-17
Add options to delete data and git branches

rustaceanrob:
When branches are merged into master, most of the time there is no point in holding on to them locally. There is also a directory that stores node data when the examples sync. It can be helpful to remove this and resync from scratch when experimenting. Also if people want to switch toolchains they might want to delete the lockfile and rebuild it.


nyonson (approved):
ACK f84996430b90773036fe31c2a9130c1d0ce7c44c

PR #338, opened 2025-04-17
Merge similar node creation fns

rustaceanrob:
These two functions essentially do the same thing. Merging them into a single one here.


nyonson (approved):
ACK 5ced5f1ec0dc377ec6b8380db29ee2f3289a177b

PR #340, opened 2025-04-17
build: add minimal dependency versions test

nyonson:
Add a test that verifies the crate compiles with minimal dependency versions using cargo's -Z direct-minimal-versions flag. This keeps us honest from accidentally depending on features in a new version of a dependency without updating the manifest.

The new test revealed a dependency on tokio `^1.19` since we use the `is_finished` function.

The `-f` setting is now used in the delete lockfile recipe so that there is no error when no file, making it easier to compose.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ bitcoin = { version = "0.32.5", default-features = false, features = [
 bip324 = { version = "0.7.0", default-features = false, features = [
     "tokio",
 ] }
-tokio = { version = "1", default-features = false, features = [
+tokio = { version = "1.19", default-features = false, features = [
     "rt-multi-thread",
     "sync",
     "time",
@@ -43,7 +43,7 @@ hex = { version = "0.4.0" }
 tracing = "0.1"
 tracing-subscriber = "0.3"
 tempfile = "3"
-tokio = { version = "1.37", default-features = false, features = [
+tokio = { version = "1.19", default-features = false, features = [
     "full",
 ] }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,3 +68,12 @@ jobs:
       - uses: extractions/setup-just@v3
       - uses: actions-rust-lang/setup-rust-toolchain@v1
       - run: just test integration
+  min-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: extractions/setup-just@v3
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: nightly
+      - run: just test min-versions
--- a/justfile
+++ b/justfile
@@ -11,7 +11,7 @@ check:
   cargo clippy --all-targets -- -D warnings
   cargo check --all-features
 
-# Run a test suite: unit, integration, features, msrv, or sync.
+# Run a test suite: unit, integration, features, msrv, min-versions, or sync.
 test suite="unit":
   just _test-{{suite}}
 
@@ -36,6 +36,11 @@ _test-features:
   cargo test --lib --no-default-features
   cargo test --lib --no-default-features --features rusqlite,filter-control 
 
+# Test that minimum versions of dependency contraints are still valid.
+_test-min-versions:
+  just _delete-lockfile
+  cargo +nightly check --all-features -Z direct-minimal-versions
+
 # Check code with MSRV compiler.
 _test-msrv:
   # Handles creating sandboxed environments to ensure no newer binaries sneak in.
@@ -54,7 +59,7 @@ _delete-data:
   rm -rf light_client_data
 
 _delete-lockfile:
-  rm Cargo.lock
+  rm -f Cargo.lock
 
 _delete-branches:
   git branch --merged | grep -v \* | xargs git branch -d


Cargo.toml
```diff
@@ -18,7 +18,7 @@ bitcoin = { version = "0.32.5", default-features = false, features = [
 bip324 = { version = "0.7.0", default-features = false, features = [
     "tokio",
 ] }
-tokio = { version = "1", default-features = false, features = [
+tokio = { version = "1.19", default-features = false, features = [
```
rustaceanrob, line 21: I don't know entirely how semver applies for libraries with stable major versions, but shouldn't this library allow for switching between minor versions? If I recall that would require an additional `1.19^`?

To be honest I don't think this is necessarily within scope for this crate. If it builds for MSRV, then the consumer should use the `tokio` version that is most compatible to their release and toolchain - at minimum 1.19, but who cares after that. I would prefer to guarantee the crate builds for _a_ minor version of `tokio`, not sure it matters entirely which one?

nyonson: > the consumer should use the `tokio` version that is most compatible to their release and toolchain - at minimum 1.19, but who cares after that

That is exactly what this constraint does. In Rust, writing a version as "1.19" in the manifest is equivalent as "^1.19.0". So "upgrade minor and patch versions, not major, and use at least version 1.19.0".

For our crate, it needs 1.19.0 because we are using functionality that was introduced in that minor version. Like you said, we want to give the consumer the most flexibility to use what ever version they want, but it must be greater than 1.19.0. The other option obviously being to refactor the code to only use tokio features available in 1.0.0, but that is a larger undertaking and I dunno if worth it.

rustaceanrob: > That is exactly what this constraint does. In Rust, writing a version as "1.19" in the manifest is equivalent as "^1.19.0".

Are you sure? I have seen the carrot syntax in `toml` before. Just want to make sure something doesn't get released that has a pointlessly strict requirement. If possible a local test to back it up would be nice.

nyonson: Ok, so I appear to be right about the implicit `^` default based on [the cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), but it is more complicated than I figured.

I ran some tests locally to be sure. If you run an aggressive `cargo update` you can see how the `1.38.2` version is used. It tops out for us due to the MSRV setting, but a consuming crate with no MSRV could go higher.

```
[mercury4 ./kyoto feat/min-versions-test]$ cargo update
    Updating crates.io index
     Locking 153 packages to latest Rust 1.63.0 compatible versions
      Adding addr2line v0.21.0 (requires Rust 1.65)
      Adding backtrace v0.3.69 (available: v0.3.74, requires Rust 1.65.0)
      Adding corepc-node v0.6.1 (available: v0.7.0)
      Adding flate2 v1.0.35 (available: v1.1.1, requires Rust 1.67.0)
      Adding once_cell v1.20.3 (available: v1.21.3, requires Rust 1.65)
      Adding r-efi v5.2.0 (requires Rust 1.68)
      Adding ring v0.17.12 (available: v0.17.14, requires Rust 1.66.0)
      Adding rusqlite v0.31.0 (available: v0.34.0)
      Adding tokio v1.38.2 (available: v1.44.2, requires Rust 1.70)
      Adding tokio-macros v2.3.0 (available: v2.5.0, requires Rust 1.70)
      Adding wasi v0.10.0+wasi-snapshot-preview1 (available: v0.10.2+wasi-snapshot-preview1)
```
And here when running the direct-minimal versions test it grabs the 1.19.0 floor.

```
[mercury4 ./kyoto feat/min-versions-test]$ cargo +nightly update -Z direct-minimal-versions
    Updating crates.io index
     Locking 148 packages
      Adding corepc-node v0.6.1 (available: v0.7.0)
      Adding flate2 v1.0.35 (available: v1.1.1, requires Rust 1.67.0)
      Adding hex v0.4.0 (available: v0.4.3)
      Adding once_cell v1.20.3 (available: v1.21.3, requires Rust 1.65)
      Adding ring v0.17.12 (available: v0.17.14, requires Rust 1.66.0)
      Adding rusqlite v0.31.0 (available: v0.34.0)
      Adding tempfile v3.0.2 (available: v3.19.1)
      Adding tokio v1.19.0 (available: v1.38.2)
      Adding tracing v0.1.0 (available: v0.1.41)
      Adding tracing-subscriber v0.3.0 (available: v0.3.19)
      Adding wasi v0.10.0+wasi-snapshot-preview1 (available: v0.10.2+wasi-snapshot-preview1)
```

rustaceanrob: I am mostly wondering then what this does for people besides those that care to inspect the `toml`. If the MSRV is 1.63.0 and the CI guarantees that builds, why does it matter _which_ `tokio` version is in the `toml`? 

Also I'm not sure this test really pinpointed what I was going for. I meant if a user made a project and brought in `tokio = 1.42` and `kyoto-cbf`, then built on stable, what would the resulting tokio version(s) be? I assume from this test that it would be 1.42 but I might still do this out of curiosity.

rustaceanrob: Alright yeah seems to work as expected. My original comment still stands, not sure I get the point. But I guess doesn't hurt to check what features are being used from `tokio`.

rustaceanrob:
ACK 27c32c8aa2500657155d3963c3c00c7272728858

Cargo.toml
nyonson: I think its good for us to be aware of actual dependency constraints we are exporting so we can help the consumers, be as easy to use as possible. If we keep the constraint at `1.0.0` and a consumer for some reason is tied to tokio `1.0.5` in their crate, the dependency resolver would look at kyoto and think "everything checks out". The consumer then tries to build their code and it explodes with confusing "no `is_finished` function is defined" errors.

PR #341, opened 2025-04-18
Pre-release clean up

rustaceanrob:
Before the next release it's best to take care of some housekeeping things. 
- It's been discussed in the past how the `README.md` code is hard to maintain, so I decided to remove it here and instead link to the Signet example. That example is well documented and up to date with the API. I also removed an old "checklist" markdown that was used for my personal development guide, but it is better to make Github issues for any remaining bugs and improvements
- Add the mainnet checkpoint required to compute the first difficulty adjustment
- DNS had a strange error I added that would make it fail if there were not enough peers. I think I was estimating that if a cache was poisoned it would respond with a few malicious peers. This is probably wrong as an attacker would likely spam with as many peers as possible. By removing the error case, DNS over Testnet4 works reliably and it didn't make much sense to have that error to begin with
- Added Testnet4 DNS seeds


nyonson (approved):
ACK 0b2b76b293a7282f2f6ead02c1a00fe92fe94a15

PR #342, opened 2025-04-18
Avoid requesting duplicate blocks

rustaceanrob:
The node may request filters of multiple peers to speed up the filter syncing process. If a filter is checked and the desired block is fetched quickly, the current code will request that block again if the filter is seen again (from a super laggy peer). Instead, only add blocks to the request queue for the filters we haven't already checked

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -537,6 +537,9 @@ impl<H: HeaderStore> Chain<H> {
 
         #[cfg(not(feature = "filter-control"))]
         if !self.block_queue.contains(&filter_message.block_hash)
+            && !self
+                .header_chain
+                .is_filter_checked(&filter_message.block_hash)
             && filter
                 .contains_any(self.scripts.iter())
                 .map_err(CFilterSyncError::Filter)?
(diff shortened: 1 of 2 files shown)


nyonson:
Just curious, have you actually seen this happen? I am still mapping out all the flows, but feels like this sorta thing could happen in a few spots. Don't have a feel for how big of a concern it is though.

Still trying to think if there is any downside, worst case scenario for this change.

nyonson:
> Not sure there is much downside since this check is constant time.

Hadn't considered performance, just some sort of terrible failure case where a block is never retrieved because the filter gets marked as checked. Not sure I have fully thought it through, but it looks like it gets marked as checked before the block is successfully retrieved right? Like, the block work has only been queued up and might fail somehow?

rustaceanrob:
The `BlockQueue` methods are infallible because it's just a simple queue that keeps recommending a block until a peer is sends it. The only way to remove a hash from the queue is to receive the block (in `check_send_block`). There isn't a failure case once the block is added to the queue AFICT

Even so, the filter is marked as checked once it is tested for matches already, this patch doesn't change any logic there

nyonson (approved):
ACK 8056d1274c2111592da6ea1468d075df142172c6

PR #343, opened 2025-04-19
Fix the `filter-control` API

rustaceanrob:
- Changes in `chain` were missed when considering duplicate filters. Removing duplicate emissions is desirable for the managed filter workflow as well as the usual flow. #342 should've caught this.

- `Filter::match_any` doesn't need to be mutable.

- It should also be possible to consume a filter and get raw bytes, perhaps to save it to disk. When loading from disk, the `BlockFilter` struct from `bitcoin` is exposed to check it for matches. Leaving `chain::Filter` private for now in case the API needs discussion.

- Implemented `Ord` on `IndexedFilter` because they can be ordered by height

- Also removed a struct I thought might be useful at the beginning of the project but hasn't proved interesting

### DIFF
--- a/example/managed.rs
+++ b/example/managed.rs
@@ -77,7 +77,7 @@ async fn main() {
             event = event_rx.recv() => {
                 if let Some(event) = event {
                     match event {
-                        Event::IndexedFilter(mut filter) => {
+                        Event::IndexedFilter(filter) => {
                             let height = filter.height();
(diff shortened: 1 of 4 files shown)


rustaceanrob:
Pretty uncontroversial. @nyonson let me know if you have any questions or push up a patch if you are unhappy about something.

PR #348, opened 2025-04-22
Fetch the `Network` from the builder

rustaceanrob:
If the builder is pass around a program, it may be useful to validate the node is being build for the expected network.


nyonson (approved):
ACK dad1008677a30b99d17cfb4df10206770f5c45e0

PR #351, opened 2025-04-27
Remove `await` points when syncing CF Headers/Filters

rustaceanrob:
With the new assumptions about `BlockTree`, such that every header will be available in memory when syncing CFHeaders and CFilters, there is no need to check the database for heights/hashes that are not present in the `BlockTree`. This removes the need for `await` points in most functions. The `send_chain_update` function is called by `node`, when processing CFHeader and CFilter messages.

This starts towards a cleaner split between sync/async methods and has seemingly increased performance on my end. Of course the cost is all the relevant headers must be stored in the `BlockTree` for a sync, but for most cases, even a recovery of 2/3 years can easily be held in RAM by most phones.

- There is a small regression in debug information where "Found a script at block" had to be removed, but it may be added back more intelligently later.
- I have a new nvim plugin to fix typos when formatting a file, so there's some random typo fixes in here

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -8,7 +8,7 @@ use std::{
 use bitcoin::{
     block::Header,
     p2p::message_filter::{CFHeaders, CFilter, GetCFHeaders, GetCFilters},
-    Block, BlockHash, Network, ScriptBuf, TxOut,
+    Block, BlockHash, Network, ScriptBuf,
 };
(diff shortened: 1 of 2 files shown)


nyonson (approved):
ACK 1a4085effe9f7f1a5bcc2ac45922bfb6aff716c1

Yep, very in line with what I was thinkin about.

PR #353, opened 2025-04-29
Use an extension trait to define what a `Height` is

rustaceanrob:
Instead of trying to figure out if a height should be `u32`, `u64` or `bitcoin::absolute::Height`, all the `graph::Height` wrapper defines is some safety rails. This is typically the use for extension traits. Instead of trying to propagate a new type around the crate, this trait can be used to define useful methods when we want to consider a `u32` as a height. In particular, we mostly want to know if a height is a difficulty adjustment, or when the last difficulty adjustment was. 

Other users could _maybe_ be interested in using this for their application logic? I have it marked as `pub` for now.

### DIFF
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -1,4 +1,4 @@
-//! Strucutres and checkpoints related to the blockchain.
+//! Structures and checkpoints related to the blockchain.
 //!
 //! Notably, [`checkpoints`] contains known Bitcoin block hashes and heights with significant work, so Kyoto nodes do not have to sync from genesis.
 pub(crate) mod block_queue;
@@ -16,7 +16,10 @@ pub(crate) mod header_batch;
 use std::collections::HashMap;
 
 use bitcoin::hashes::{sha256d, Hash};
-use bitcoin::{bip158::BlockFilter, block::Header, BlockHash, FilterHash, FilterHeader, ScriptBuf};
+use bitcoin::{
+    bip158::BlockFilter, block::Header, params::Params, BlockHash, FilterHash, FilterHeader,
+    ScriptBuf,
+};
 
 use crate::network::PeerId;
 
@@ -188,9 +191,41 @@ impl HeightMonitor {
     }
 }
 
+// Attributes of a height in the Bitcoin blockchain.
+trait HeightExt: Clone + Copy + std::hash::Hash + PartialEq + Eq + PartialOrd + Ord {
+    fn increment(&self) -> Self;
+
+    fn from_u64_checked(height: u64) -> Option<Self>;
+
+    fn is_adjustment_multiple(&self, params: impl AsRef<Params>) -> bool;
+    #[allow(unused)]
+    fn last_epoch_start(&self, params: impl AsRef<Params>) -> Self;
+}
+
+impl HeightExt for u32 {
+    fn increment(&self) -> Self {
+        self + 1
+    }
+
+    fn is_adjustment_multiple(&self, params: impl AsRef<Params>) -> bool {
+        *self as u64 % params.as_ref().difficulty_adjustment_interval() == 0
+    }
+
+    fn from_u64_checked(height: u64) -> Option<Self> {
+        height.try_into().ok()
+    }
+
+    fn last_epoch_start(&self, params: impl AsRef<Params>) -> Self {
+        let diff_adjustment_interval = params.as_ref().difficulty_adjustment_interval() as u32;
+        let floor = self / diff_adjustment_interval;
+        floor * diff_adjustment_interval
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bitcoin::Network;
 
     #[test]
     fn test_height_monitor() {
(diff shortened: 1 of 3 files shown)


nyonson:
I am still a little new to the subtleties between newtypes and type extensions ([went on a little journey](https://blog.yonson.dev/log/2025-04-vol11/)), but I believe the trade-off being made here is simply type safety for flexibility. In bitcoin land, I want the type checker to ideally double check as much as possible, even at the cost of some extra boilerplate, so this feels dicey. But I'd like to get a better understanding of the cost in this scenario.

The big cost of the newtype is boilerplate, but are there ways to bring that down to some acceptable level so that we can still have the type checker in our corner? Combo-ing the type extension with a type alias is a cool approach, gives some nice documentation, but the type checker is still out of the loop and won't help enforce any safety rails. So leaves me uneasy. 

Some things I am thinking on,

1. You mention the pain of propagating a newtype around the crate, but don't we still kinda have to do this with the extension + alias too? At least have to make them public and import them everywhere. Just want to make sure I understand the cost here.
2. If we stick with a newtype, I don't think it has to be exposed to the consumers, so at least the cost is only on the library. We could keep the external API's as `u32`, or maybe more flexible in the future, have the new type implement `impl From<u32> for BlockHeight` and make the external functions super flexible with `fn process_height<H: Into<BlockHeight>>(height: H) {...}`.
3. How many of the boilerplate concerns can be iron'd out with `From` impls? And maybe a few `ops` or a full on `Deref` (although this one sounds pretty powerful)? Is it possible to get it easy enough to keep the type checker?

I looked over at the rust-bitcoin version for inspiration, which lives over in the new `units` crate now, [src/locktime/absolute.rs](https://github.com/rust-bitcoin/rust-bitcoin/blob/a9ddac178d937dce434ff4f9e8831055983a379e/units/src/locktime/absolute.rs#L29). They roll with a newtype there which is worth noting, but it has a lot of consensus serialization attached to it which I believe will always be out of scope for kyoto, so probably not worth using directly.

src/chain/mod.rs
```diff
@@ -188,9 +191,52 @@ impl HeightMonitor {
...
+        height.try_into().ok()
+    }
+
+    fn last_epoch_start(&self, params: impl AsRef<Params>) -> Self {
+        let diff_adjustment_interval = params.as_ref().difficulty_adjustment_interval() as u32;
+        let floor = self / diff_adjustment_interval;
+        floor * diff_adjustment_interval
+    }
+
+    fn to_u32(self) -> u32 {
```
nyonson, line 231: `to_u32` looks pretty weird on a `for u32`. I guess `HeightExt` could be implemented on other types in the future, but feels like a stretch atm.

src/chain/mod.rs
```diff
@@ -188,9 +191,52 @@ impl HeightMonitor {
...
+
+    /// The latest difficulty epoch period start height, relative to this one.
+    fn last_epoch_start(&self, params: impl AsRef<Params>) -> Self;
+
+    /// Cast a height to `u32`.
+    fn to_u32(self) -> u32;
+}
+
+impl HeightExt for u32 {
+    fn increment(&self) -> Self {
```
nyonson, line 213: Feels like the safety of this method is not quite as powerful now since the u32 ops are readily available.

src/chain/mod.rs
rustaceanrob: Good catch, this could actually restrict what we want.

src/chain/mod.rs
rustaceanrob: Definitely. I think for me it is more about code legibility as I don't like how `height + 1` reads compared to `height.increment()`, but it is only an encouragement with the extension trait. I'm going to leave it for the sake of less refactor surface

rustaceanrob:
> You mention the pain of propagating a newtype around the crate, but don't we still kinda have to do this with the extension + alias too? At least have to make them public and import them everywhere. Just want to make sure I understand the cost here.

Updated in dc9ef061233210f0ec89b21f932b917c76fe290f, I generally agree that trait should be private and local to the `chain` module. Bringing in `HeightExt` is a simple import, whereas the `Height` abstraction requires a `Height::new`, execute height specific operations, then cast back to a primitive. This is exactly what the extension trait serves without the conversion steps.

> If we stick with a newtype, I don't think it has to be exposed to the consumers

If we want to use the benefits of the wrapper that forces a strict API, the type will slowly creep around the crate, to the point where every public API will be calling `to_u32`. That is why I am preferring the all-or-nothing approach, as there are unanswered questions with a wrapper. Would it be possible to use `RangeBounds` required for the `HeaderStore`? Would there be a plan to use a `bitcoin` wrapper in the future? If so, that would require the extension trait anyway. If there is an intermediate ground where the wrapper is used in some places and not others, then all of the benefits of the wrapper can be casted away, and the types have to be explicitly converted to/from, which makes code harder to maintain and reason about. 

> at least the cost is only on the library

I want to reduce my workload as much as possible

> How many of the boilerplate concerns can be iron'd out with From impls? And maybe a few ops or a full on Deref

This is boilerplate imo.

> I looked over at the rust-bitcoin version for inspiration, which lives over in the new units crate now, [src/locktime/absolute.rs](https://github.com/rust-bitcoin/rust-bitcoin/blob/a9ddac178d937dce434ff4f9e8831055983a379e/units/src/locktime/absolute.rs#L29).

It implements all of the +, -, /, *, so the only point of this type is to not add a `Time` with `Height` or something. There's no arithmetic difference between `Height` and `u32`.

rustaceanrob:
Gonna strong-arm this one cause this is closer to an application than a library in many respects, so decision paralysis over these sorts of things holds up the development process. I think this refactor is best left for `bitcoin 1.0.0` when it makes sense to get everyone in the ecosystem on the same abstractions. Currently everyone higher up the stack uses `u32`

nyonson:
> Would it be possible to use `RangeBounds` required for the `HeaderStore`?

Hm, that one does appear to be a bit of doozy. I can see not wanting to overly invest in newtype if there are uncertainties in the future.

> > I looked over at the rust-bitcoin version for inspiration, which lives over in the new units crate now, [src/locktime/absolute.rs](https://github.com/rust-bitcoin/rust-bitcoin/blob/a9ddac178d937dce434ff4f9e8831055983a379e/units/src/locktime/absolute.rs#L29).
> 
> It implements the +, - so the only point of this type is to not add a `Time` with `Height` or something. There's no arithmetic difference between `Height` and `u32`.

It implements those operators? I don't see that anywhere...and I missing something obvious? I think its `derive`'s only give it equality and order operators.

PR #354, opened 2025-04-29
Load relevant block headers into memory at startup

rustaceanrob:
Based on #353 to show where I am going with that PR.

Currently there is a strange bit of code to handle the case of an unknown previous hash when accepting a block header. We check the database if the hash is present there, and then reinitialize the `BlockTree` from that header if present in the database. Because the `BlockTree` holds every header in memory now, a better approach is to just load in headers that are 7 blocks in the past in at the start. Better yet, we can also load in the header necessary to compute the next difficulty adjustment. In this PR, I attempt to load as far back in the history as possible to gracefully handle reorganizations and difficulty adjustments. 

One caveat is that this would be ignoring where the user requested to start their scan from. Re-emitting blocks is [forbidden in some cases](https://docs.rs/lightning/0.1.2/src/lightning/ln/channelmanager.rs.html#10835) in higher level LN implementations, so the start height must be adhered to. To compensate for this, I added an `assumed_checked_to` method on `BlockTree`, that marks each filter in the chain loaded from disk as checked, up to the height the user requested. Note that, because the filter messages are constructed by traversing the block tree, reorganizations will still check filters below this height if necessary.

This removes the need to do any further loading after the initialization in `Node::run`

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -17,9 +17,9 @@ use super::{
     cfheader_batch::CFHeaderBatch,
     checkpoints::{HeaderCheckpoint, HeaderCheckpoints},
     error::{BlockScanError, CFHeaderSyncError, CFilterSyncError, HeaderSyncError},
-    graph::{AcceptHeaderChanges, BlockTree, HeaderRejection, Tip},
-    CFHeaderChanges, Filter, FilterHeaderRequest, FilterRequest, FilterRequestState, HeightMonitor,
-    PeerId,
+    graph::{AcceptHeaderChanges, BlockTree, HeaderRejection},
+    CFHeaderChanges, Filter, FilterHeaderRequest, FilterRequest, FilterRequestState, HeightExt,
+    HeightMonitor, PeerId,
 };
 #[cfg(feature = "filter-control")]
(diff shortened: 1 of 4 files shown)


src/chain/chain.rs
```diff
@@ -139,13 +109,35 @@ impl<H: HeaderStore> Chain<H> {
...
+    // reorganizations to be handled gracefully.
     pub(crate) async fn load_headers(&mut self) -> Result<(), HeaderPersistenceError<H::Error>> {
-        let loaded_headers = self
-            .db
-            .lock()
-            .await
-            .load(self.header_chain.height() + 1..)
+        let mut db = self.db.lock().await;
+        // The original height the user requested a scan after
+        let scan_height = self.header_chain.height();
```
nyonson, line 117: I found it a little confusing to bounce between the `header_chain` to then re-init the `header_chain`, wasn't clear at first why that initial state is important. Maybe can be iron'd out in the future where the config is used to initialize it just once to the expected spot.

nyonson (approved):
ACK 46c42cd4c0d80544239e24d2256e052de68b76ab

PR #356, opened 2025-04-29
feat/mempool: example service that emits mempool and blockchain data via ws

matthiasdebernardini:


rustaceanrob:
Hey this is super neat. I'm glad you took some time to experiment with this.

Because I am already stretched thin on the development side, I would rather not bring in any new dev dependencies, as they add compilation time both locally and in CI. For examples as well, the guarantee of the crate is they will be maintained, and I am not sure I would be able to uphold that for this example.

For what its worth I can leave this up for others to check out if their curious, however I won't be merging this at the moment

matthiasdebernardini:
All good! I'll link it to the issue in case anyone searches for it.

PR #357, opened 2025-04-30
Stage and flush changes, instead of writing for each header

rustaceanrob:
Currently every new header is being written to disk one by one. When catching up to the chain tip, these messages are 2,000 headers long. By saving the writes until the end of the message processing, it's essentially a 2,000x speed up in sync time because the writes are by far the most expensive operation. 

This PR introduces a two steps to writing. First the database stages changes, which will likely be an in-memory operation, and should be both blocking and infallible. Next the database writes the changes to disk, which may be asynchronous. For most cases, this preserves the nice properties of evaluating one block header at a time with `stage`, but separates the performance hit into a different step.

Given the mutability rules of Rust, I made a boolean flag to catch if there was a reorg that occurred, and to reset the compact filter message queue if there was one. This allows for a single lock on the database, which is then dropped once the changes are written, and allows for `self` to be mutable once more to clear the message queue if required.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -100,15 +100,6 @@ impl<H: HeaderStore> Chain<H> {
         }
     }
 
-    // Write the chain to disk
-    pub(crate) async fn write_changes(&mut self, changes: BlockHeaderChanges) {
-        if let Err(e) = self.db.lock().await.write(changes).await {
-            self.dialog.send_warning(Warning::FailedPersistence {
-                warning: format!("Could not save headers to disk: {e}"),
-            });
-        }
-    }
-
     // Load in headers, ideally allowing the difficulty adjustment to be audited and
     // reorganizations to be handled gracefully.
     pub(crate) async fn load_headers(&mut self) -> Result<(), HeaderPersistenceError<H::Error>> {
@@ -189,6 +180,8 @@ impl<H: HeaderStore> Chain<H> {
         // We check first if the peer is sending us nonsense
         self.sanity_check(&header_batch)?;
         let next_checkpoint = self.checkpoints.next().copied();
+        let mut db = self.db.lock().await;
+        let mut reorg_occured = false;
         for header in header_batch.into_iter() {
             let changes = self.header_chain.accept_header(header);
             match changes {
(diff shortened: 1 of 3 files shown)


src/chain/chain.rs
```diff
@@ -268,6 +259,15 @@ impl<H: HeaderStore> Chain<H> {
                 },
             }
         }
+        if let Err(e) = db.write().await {
+            self.dialog.send_warning(Warning::FailedPersistence {
+                warning: format!("Could not save headers to disk: {e}"),
+            });
+        }
+        drop(db);
```
nyonson, line 267: Is this explicit `drop` required due to async things? Figured the compiler would be smart enough to drop it here for us.

nyonson (approved):
ACK 044e9fbb1252572d540715ac53a9dddedc6f0fe9

src/db/traits.rs
```diff
@@ -12,15 +12,17 @@ use super::{BlockHeaderChanges, PersistedPeer};
...
     ) -> FutureResult<'a, BTreeMap<u32, Header>, Self::Error>;
 
-    /// Write any changes to the backend as new headers are found, potentially caching before
-    /// writing to disk.
-    fn write(&mut self, changes: BlockHeaderChanges) -> FutureResult<(), Self::Error>;
+    /// Stage changes to the chain to be written in the future.
+    fn stage(&mut self, changes: BlockHeaderChanges);
+
+    /// Commit the changes by writing them to disk.
+    fn write(&mut self) -> FutureResult<(), Self::Error>;
```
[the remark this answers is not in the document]
rustaceanrob: I was just reading about `Pin`, I think the trait will probably still be syntax sugar for `Pin<Box<dyn Future>>`? Futures are self-referential so that puts the compiler in a hard spot.

src/chain/chain.rs
rustaceanrob: This is with what's going on with the `Mutex`. The lock is held for the current scope, so if we tried to take a lock again, we would deadlock. It's one of those things the compiler can't really help with, but what it _does_ know is that we can't borrow `&mut self` twice. This is just to tell the compiler "I am done with the lock" without making weird brackets with no expression

PR #359, opened 2025-05-02
fix(network): drop dynamic dispatch on read path

nyonson:
Breaking this refactor up into separate steps since it is getting pretty hairy. The overall goal is to get rid of dynamic dispatch in the network module. Starting here with the read path. The write path can be done in a separate patch since its impl is slightly different. I have some ideas for general clean up as well (e.g. make use of the lower level bip324 types), but going to save those till after to keep the changesets clearer.

* Dropping the Stream{Reader/Writer} types to avoid dynamic dispatch on the message read path.
* Have opted to pass around a concrete TcpStream more often and only splitting it when needed.
* The read implementations are now an enum avoiding any major modifications for now.

### DIFF
--- a/src/network/mod.rs
+++ b/src/network/mod.rs
@@ -9,11 +9,7 @@ use bitcoin::{
     p2p::{address::AddrV2, message::CommandString, Magic},
 };
 use socks::create_socks5;
-use tokio::{
-    io::{AsyncRead, AsyncWrite},
-    net::TcpStream,
-    time::Instant,
-};
+use tokio::{net::TcpStream, time::Instant};
 
 use error::PeerError;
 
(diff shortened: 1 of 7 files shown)


rustaceanrob:
Awesome, bit of spring cleaning.

> Dropping the Stream{Reader/Writer} 

For future reference: these were used for `arti` integration, which did not use `TcpStream`, and didn't serve much engineering purpose. Choosing the `TcpStream` directly _will_ cut out web-transport if it becomes available in the future, but I think that is an acceptable tradeoff. The private tradeoffs of using compact block filters are obscured when it comes to the browser. Limiting support to desktop, mobile, and potentially embedded over small linux distros is a good project policy.

rustaceanrob:
ACK 38c17bc903d98d0239a45c75dab606bdf46c83de

PR #360, opened 2025-05-05
Improve the peer timeout API

rustaceanrob:
First commit:
- Consolidate the separate fields of the config that determined timeout values into the `PeerTimeoutConfig`, which already existed. Moves a couple constants to the `network` module so `PeerTimeoutConfig` can implement `Default`.

Second commit:
- Uses the `impl Into<Duration>` syntax, because a user should be free to swap `std::time::Duration` and `tokio::time::Duration`
- Adds a new handshake timeout field to `PeerTimeoutConfig`, which is passed to the `connect` method as an argument.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -165,7 +165,7 @@ impl NodeBuilder {
     ///
     /// If none is provided, a timeout of 5 seconds will be used.
     pub fn response_timeout(mut self, response_timeout: Duration) -> Self {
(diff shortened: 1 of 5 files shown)


rustaceanrob:
@nyonson I am going to go ahead and merge this, but tagging you here so you can see the diff. Hopefully doesn't mess with your work to remove dynamic dispatch.

PR #363, opened 2025-05-05
fix(network): remove write path dynamic dispatch

nyonson:
Using an enum for static dispatch clarity and performance.


src/network/peer.rs
```diff
@@ -91,13 +84,16 @@ impl Peer {
                 self.dialog.send_warning(Warning::CouldNotConnect);
             }
             let (decryptor, encryptor) = handshake_result?;
-            let message_mutex: MutexMessageGenerator =
-                Mutex::new(Box::new(V2OutboundMessage::new(self.network, encryptor)));
+            let message_mutex = Mutex::new(MessageGenerator::V2 {
```
rustaceanrob, line 87: No more need for a mutex in either of these paths. We can return the `MessageGenerator` directly and take `&mut` references in both of the `select` arms. Here's the diff over your current patch: 

```diff
diff --git a/src/network/peer.rs b/src/network/peer.rs
index 565bf2a..0fce00b 100644
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -1,5 +1,5 @@
 extern crate tokio;
-use std::{collections::HashMap, ops::DerefMut, sync::Arc, time::Duration};
+use std::{collections::HashMap, sync::Arc, time::Duration};

 use bip324::{AsyncProtocol, PacketReader, PacketWriter, Role};
 use bitcoin::{p2p::ServiceFlags, Network, Transaction, Wtxid};
@@ -7,10 +7,7 @@ use tokio::{
     io::{AsyncRead, AsyncWrite, AsyncWriteExt},
     net::TcpStream,
     select,
-    sync::{
-        mpsc::{self, Receiver, Sender},
-        Mutex,
-    },
+    sync::mpsc::{self, Receiver, Sender},
     time::Instant,
 };

@@ -69,7 +66,7 @@ impl Peer {
         let (tx, mut rx) = mpsc::channel(32);
         let (mut reader, mut writer) = connection.into_split();
         // If a peer signals for V2 we will use it, otherwise just use plaintext.
-        let (message_mutex, mut peer_reader) = if self.services.has(ServiceFlags::P2P_V2) {
+        let (mut outbound_messages, mut peer_reader) = if self.services.has(ServiceFlags::P2P_V2) {
             let handshake_result = tokio::time::timeout(
                 Duration::from_secs(HANDSHAKE_TIMEOUT),
                 self.try_handshake(&mut writer, &mut reader),
@@ -84,22 +81,20 @@ impl Peer {
                 self.dialog.send_warning(Warning::CouldNotConnect);
             }
             let (decryptor, encryptor) = handshake_result?;
-            let message_mutex = Mutex::new(MessageGenerator::V2 {
+            let message = MessageGenerator::V2 {
                 network: self.network,
                 encryptor,
-            });
+            };
             let reader = Reader::new(MessageParser::V2(reader, decryptor), tx);
-            (message_mutex, reader)
+            (message, reader)
         } else {
-            let message_mutex = Mutex::new(MessageGenerator::V1 {
+            let message = MessageGenerator::V1 {
                 network: self.network,
-            });
+            };
             let reader = Reader::new(MessageParser::V1(reader, self.network), tx);
-            (message_mutex, reader)
+            (message, reader)
         };

-        let mut message_lock = message_mutex.lock().await;
-        let outbound_messages = message_lock.deref_mut();
         let message = outbound_messages.version_message(None)?;
         self.write_bytes(&mut writer, message).await?;
         self.message_counter.sent_version();
@@ -134,7 +129,7 @@ impl Peer {
                     if let Ok(peer_message) = peer_message {
                         match peer_message {
                             Some(message) => {
-                                match self.handle_peer_message(message, &mut writer, outbound_messages).await {
+                                match self.handle_peer_message(message, &mut writer, &mut outbound_messages).await {
                                     Ok(()) => continue,
                                     Err(e) => {
                                         match e {
@@ -153,7 +148,7 @@ impl Peer {
                 node_message = self.main_thread_recv.recv() => {
                     match node_message {
                         Some(message) => {
-                            match self.main_thread_request(message, &mut writer, outbound_messages).await {
+                            match self.main_thread_request(message, &mut writer, &mut outbound_messages).await {
                                 Ok(()) => continue,
                                 Err(e) => {
                                     match e {
```

src/network/outbound_messages.rs
```diff
@@ -19,218 +19,271 @@ use bitcoin::{
...
     }
 
-    fn verack(&mut self) -> Result<Vec<u8>, PeerError> {
-        let data = RawNetworkMessage::new(self.network.magic(), NetworkMessage::Verack);
-        Ok(serialize(&data))
+    pub(crate) fn verack(&mut self) -> Result<Vec<u8>, PeerError> {
+        match self {
+            MessageGenerator::V1 { network } => {
+                let data = RawNetworkMessage::new(network.magic(), NetworkMessage::Verack);
+                Ok(serialize(&data))
```
rustaceanrob, line 55: I think if we move the declaration of the message above the `match` statement, we can use an internal function to determine the path to serializing, so each function could just be:

```
pub(crate) fn verack(&mut self) -> Result<Vec<u8>, PeerError>> {
    let msg = NetworkMessage::Verack;
    self._write(msg)
}

fn _write(&mut self, msg: NetworkMessage) -> Result<Vec<u8>, PeerError>> {
   /// match impl here
}

rustaceanrob:
Saved you the trouble in 7b30217bed4f8a84f1e80de0060d4859152920a5

Thanks man

PR #372, opened 2025-05-08
fix: remove dead code in network module

nyonson:


rustaceanrob:
ACK 89d9548596e7edd1b5a04923f9ab488a24d45845

Yeah please rebase, pretty much all related. Thanks man, the cleanup continues.

rustaceanrob:
Oh shit actually I just remembered the `bip324` implementation is significantly different in terms of reading off bytes. I want to be careful about performance regressions and might want to take some time to benchmark this on mainnet. In the Kyoto implementation, cancellation is irrelevant because the reader is on its own task, so if the cancel-safety state-machine is an overhead I would probably NACK this

rustaceanrob:
I thought about it more and I think the primary reason I am wary is, even if equally performant, the `bip324` implementation is extremely hard to decipher and would be nearly impossible to debug. For the implementation we have:

```rust
        loop {
            match &mut self.state {
                DecryptState::ReadingLength {
                    length_bytes,
                    bytes_read,
                } => {
                    while *bytes_read < 3 {
                        *bytes_read += buffer.read(&mut length_bytes[*bytes_read..]).await?;
                    }

                    let packet_bytes_len = self.packet_reader.decypt_len(*length_bytes);
                    let packet_bytes = vec![0u8; packet_bytes_len];
                    self.state = DecryptState::ReadingPayload {
                        packet_bytes,
                        bytes_read: 0,
                    };
                }
                DecryptState::ReadingPayload {
                    packet_bytes,
                    bytes_read,
                } => {
                    while *bytes_read < packet_bytes.len() {
                        *bytes_read += buffer.read(&mut packet_bytes[*bytes_read..]).await?;
                    }

                    let payload = self.packet_reader.decrypt_payload(packet_bytes, None)?;
                    self.state = DecryptState::default();
                    return Ok(payload);
                }
            }
        }
```

Which is in contrast to:

```rust
let mut len_buf = [0; 3];
let _ = stream.read_exact(&mut len_buf);
```

Although not cancel safe, it is extremely easy to reason about what is going on here. How bytes are being read off of the wire is not something I want to think about when debugging, and the current implementation does not need to be cancel safe, nor do I think it should have to be cancel safe, or else something is poorly designed. I think this should be sorted somehow in `bip324` to either ignore cancel safety or throw it through some intense fuzzing

nyonson:
> Which is in contrast to:
> 
> ```rust
> let mut len_buf = [0; 3];
> let _ = stream.read_exact(&mut len_buf).map_err(|_| PeerReadError::ReadBuffer)?;
> ```
> 
> Although not cancel safe, it is extremely easy to reason about what is going on here. How bytes are being read off of the wire is not something I want to think about when debugging. The current implementation does not need to be cancel safe, nor do I think it should have to be cancel safe, or else something is poorly designed. I think this should be sorted somehow in `bip324` to either ignore cancel safety or throw it through some intense fuzzing

This is not a 1:1 comparison since the top code is reading off the length, payload, and doing the decryption versus the bottom only reading the length, but the performance and use-ability concerns are valid. My goal is for bip324 to be incredibly rock solid and don't want to force a switch to this interface if it isn't. I'll tried to shore up any tests concerns there, but after just a brief glance, not sure how this would fail beyond the standard I/O or decryption errors...

rustaceanrob:
As a precursor, the attitude of this crate is that a `panic` could potentially result in a loss of funds, as we cannot assume what the node is being used for. Except for very obvious cases, like the system clock should be set past the unix epoch time, there are no panic conditions.

The fundamental difference is that the `bip324` code includes slice indexing by variable, which could panic. While the state-machine seems impossible to make an invalid access into the slice, I wouldn't be comfortable without confirming with fuzz testing or similar. What I meant by showing the `read_exact` example, is reproducible and predictable behavior that cannot be exploited without a simple disconnection. I am just not confident that the `bip324` implementation has that same property by looking at it with the naked eye.

As far as performance, the `read` call just calls `read` again, and is at the mercy of the runtime of when it will be executed. I am wondering if the `read_exact` implementation of TCP Stream can register with the runtime and say "only wake me if you have this many bytes ready", which would be more performant. The `tokio` implementations work very closely with the OS

Issue #367, opened 2025-05-06
Handle `AddrV2` gossip from the P2P network on a separate task


rustaceanrob:
**Describe the enhancement**  

When peers are gossiped to the node it blocks the main application loop to add them. Instead we can run a separate task with a reference counted `PeerStore` to handle gossip over a channel. `PeerMap` can hang on to a RC'd `PeerStore` as well, but will only need it to update peers or randomly select them

**Use case**  

Unblock the main thread when peers get added.

rustaceanrob:
@nyonson I can't assign you directly but I assume you can take this one up

rustaceanrob:
Closed by #373

PR #373, opened 2025-05-09
Allow `Peer` to manage gossiped peers

rustaceanrob:
Allowing the `Peer` type to add values to the database unblocks the main task when new peers are gossiped to the node. The `PeerStore` requires a static lifetime for the node to work (see `NodeBuilder`), so the restriction is only making it explicit here.

Bitcoin Core will rarely and randomly gossip one or two peers during a connection. These don't require a new task to add, so I made a threshold of 10 to actually begin a sub-task. The sub task just takes a single lock on the database and adds all the addresses gossiped. This tidies up the `PeerMap` and `Node` at the expense of making `Peer` more involved. 

Locally this does as I intended, and the header sync can occur while gossiped peers are added in the background. 

This is one possible solution to #367

### DIFF
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -17,9 +17,10 @@ use tokio::{
 use crate::{
     broadcaster::BroadcastQueue,
     channel_messages::{MainThreadMessage, PeerMessage, PeerThreadMessage, ReaderMessage},
+    db::PersistedPeer,
     dialog::Dialog,
     messages::Warning,
-    Info,
+    Info, PeerStore,
 };
 
 use super::{
@@ -33,19 +34,20 @@ use super::{
 const MESSAGE_TIMEOUT: u64 = 2;
 const HANDSHAKE_TIMEOUT: u64 = 4;
 
-pub(crate) struct Peer {
+pub(crate) struct Peer<P: PeerStore + 'static> {
     nonce: PeerId,
     main_thread_sender: Sender<PeerThreadMessage>,
     main_thread_recv: Receiver<MainThreadMessage>,
     network: Network,
     services: ServiceFlags,
     dialog: Arc<Dialog>,
+    db: Arc<Mutex<P>>,
     timeout_config: PeerTimeoutConfig,
     message_state: MessageState,
     tx_queue: Arc<Mutex<BroadcastQueue>>,
 }
 
-impl Peer {
+impl<P: PeerStore + 'static> Peer<P> {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         nonce: PeerId,
@@ -54,6 +56,7 @@ impl Peer {
         main_thread_recv: Receiver<MainThreadMessage>,
         services: ServiceFlags,
         dialog: Arc<Dialog>,
+        db: Arc<Mutex<P>>,
         timeout_config: PeerTimeoutConfig,
         tx_queue: Arc<Mutex<BroadcastQueue>>,
     ) -> Self {
@@ -64,6 +67,7 @@ impl Peer {
             network,
             services,
             dialog,
+            db,
             timeout_config,
             message_state: MessageState::new(timeout_config.response_timeout),
             tx_queue,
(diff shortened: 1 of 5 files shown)


nyonson:
Still hacking on a bit of doc tests here, but this feel generally cleaner to keep the logic in peermap instead of pulling the db down to peer: https://github.com/2140-dev/kyoto/pull/375

src/network/peer.rs
```diff
@@ -413,4 +417,52 @@ impl Peer {
             }
         }
     }
+
+    async fn add_gossiped_peers(dialog: Arc<Dialog>, db: Arc<Mutex<P>>, peers: Vec<CombinedAddr>) {
+        let len = peers.len();
+        crate::log!(dialog, format!("Adding {len} gossiped peers"));
+        if len < PEERS_GOSSIPED_TASK_THRESHOLD {
```
nyonson, line 424: This feels a bit premature optimization to me. The double-code and extra arc clones for the db and dialog. Could we just stick the db insert into the PeerMessage::Addr handler directly for now?

rustaceanrob: Not sure what you mean entirely. From a protocol point of view, there will be times where we either get one or two randomly gossiped peers, or a firehouse at the beginning of the exchange. The task should only be used when we get the initial gossip. In fact, there should probably be some state involved here such that this task is only ever called once, otherwise the peer is spamming.

rustaceanrob: Covered in [counter](https://github.com/2140-dev/kyoto/blob/master/src/network/counter.rs), but I was going to revisit rate-limiting and message counting soon

nyonson: So if the rate-limiting is gonna happen else where, which makes sense to me, could this just be simplified to just write to the db on the current task? Could maybe have a simple "max number of peers" cap (e.g. only writes the first 10 and ignores the rest), but seems straight forward.

rustaceanrob: That would be dropping a lot of peers on the floor. Prospective peers should be considered very valuable, as the proportion of nodes running CBF is relatively low. In terms of gossip, the numbers for mainnet are roughly 250-300 peers get advertised to us after we sent a `getaddr`. After that, it seems to be a sort of ping-style message, where the peer will randomly send 1-3 every minute or two. Basically there are two states, we expect a lot of peers, or we expect very little. This domain-specific component can be leveraged to choose if a new task should be created or not. 

To re-iterated on what the purpose of this change is, the main thread is blocked while adding this initial gossip every time a new peer is connected to. Moving the block to the `Peer` doesn't solve the fundamental problem that messages aren't being processed. It makes it slightly better as other peers are not blocked, but, at the limit, we can just un-block all processes with a new task. This task has a hard-limit of one task per `Peer`, or else the remote node is doing something faulty or nefarious.

This current change is not great because of the magic number logic. Instead I'm going to start introducing some new state to `Peer`, so deciding when/if to spawn a task for `addr` gossip is trivial

rustaceanrob: Removed the magic numbers in 8ab93fc3582002ff81a3ae9379d21b4b39de73d4

The `AddrGossipStages` now exists in either `NotReceived`, so the beginning of the connection, or `RandomGossip`, where we have received the response to our request for new addrs, and now we expect random and unsolicited addrs. We can optimistically spawn a new `tokio::task` in the case of `NotReceived` because this should be the fire-hose of new peers at the start of the connection. After that, we can block when adding the peers. This allocates a single extra `tokio::task` per connection in a well-defined way.

After revisiting this, it does still feel a little gross to have the `Peer` be holding this state, but it reduces bloat in `Node` and `PeerMap`, so am still thinking it is a net win.

nyonson: Yea, I think it all makes sense, especially with those numbers you posted on the usual behavior. I like the variant arms vs. the magic, definitely makes it clearer to me.

nyonson (approved):
ACK c8c301111195a8c9813f3ba67a7c4f28148698bd

PR #375, opened 2025-05-09
feature(network): handle gossip peers in peermap

nyonson:
Let the peermap handle gossip peers itself instead of bubbling those events up to the node and having them passed back down again. This keeps the node focused on node things and uninterrupted.


rustaceanrob:
Although less invasive with passing the `PeerStore` around, we have to add this channel overhead to the `Peer`. The `Peer` would be adding addresses to the channel and this task would poll for them, but if the `Peer` has the address, I'm leaning towards the attitude that the `Peer` should just add them. Some more thought has to go into #373, but I am preferring that approach so far.

PR #378, opened 2025-05-14
Unpin `bitcoin` patch version

rustaceanrob:
Cargo check seems to ignore the pinned version of `bitcoin` to `0.32.5` and builds `0.32.6`. This results in a new lint. We should be picking up patch releases from `bitcoin` anyway. Unpinning the `.5`


rustaceanrob:
Alright well now the direct minimum versions test is failing but we need to assess if that test is even useful.

PR #392, opened 2025-05-26
Use DNS as a source for initial peers

### LINKED ISSUE #368: Allow querying DNS on every sync
**Describe the enhancement**  

DNS seeders have a single job to crawl the bitcoin P2P network. When querying a seeder for peers to connect to, it is highly likely that response will contain active nodes. For a light client that does not have high uptime, some of the peer selection can be delegated to always-online servers, if the developer and user are comfortable with the tradeoffs

**Use case**  

Sync times would be greatly improved at the expensive of privacy to the DNS resolver - probably not relevant - and trust that the DNS seeder is serving honest nodes.

rustaceanrob:
Presumably, some end-users of this crate will download network data infrequently. This means the node may have to iterate through a high number of stale peers within the peer store, which ultimately leads to a poor user experience. The DNS seeders are acting as always-on network crawlers. If a developer is comfortable with trusting these seeders, they can provide high quality potential connections. In my opinion this isn't much of a privacy tradeoff, as the DNS resolver only knows the user is interested in the bitcoin network. This is, however, a significant trust tradeoff, as now the DNS seeders are used as a source-of-truth for good bitcoin nodes.

Guidance on if this feature should be recommended or not depends on some network analysis. @nyonson if your research finds that the proportion of filter-serving nodes _listening_ for inbound connections is low enough, this would be okay to promote to end-developers. On the other hand, if the network is reasonably available to serve filters, this could be used in special cases where we don't expect the node to shut down frequently.

This is controversial in general, so this will never be a default.

Closes #368

### DIFF
--- a/example/bitcoin.rs
+++ b/example/bitcoin.rs
@@ -31,6 +31,10 @@ async fn main() {
         .after_checkpoint(checkpoint)
         // The number of connections we would like to maintain
         .required_peers(2)
+        // Random resolver pulled from: https://public-dns.info/nameservers.txt
+        .dns_resolver([185, 220, 182, 179])
+        // Query the DNS resolver on every sync. Read more about the traeoffs in the documentation.
+        .force_dns_seeding()
         // Create the node and client
         .build()
         .unwrap();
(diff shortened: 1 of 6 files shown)


rustaceanrob:
Closing in favor of a less opinionated approach

PR #424, opened 2025-06-18
DNS bootstrap at the start of node lifecycle

rustaceanrob:
When trying to select a random peer from the database, if the database happens to have 0 peers available, and there are no trusted peers, then the DNS bootstrap occurs. This special logic does not need to occur when selecting a peer to connect to. Instead, the node may boostrap if necessary at the start of the process, given there are no peers available.


rustaceanrob:
I just realized this has weird implications if we have a list of trusted peers at first, and none of those nodes are available. The previous behavior would resort to DNS, but this would just make the node return an error.

PR #446, opened 2025-09-03
Remove the header store

rustaceanrob:


rustaceanrob:
I was working on a description as you commented! @thunderbiscuit

rustaceanrob:
While you're hanging out in this part of town, do you have an opinion on #445 @thunderbiscuit?

PR #556, opened 2026-03-18
Fix broadcasting to a peer who already has tx

randomlogin:
Previously if we broadcast to a peer who already has the tx it would hang indefinetely, as there is no subsequent getdata message.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use bitcoin::p2p::address::AddrV2;
 use bitcoin::p2p::ServiceFlags;
 use bitcoin::{Amount, Transaction, Wtxid};
@@ -32,9 +34,10 @@ impl Client {
         warn_rx: mpsc::UnboundedReceiver<Warning>,
         event_rx: mpsc::UnboundedReceiver<Event>,
         ntx: UnboundedSender<ClientMessage>,
+        broadcast_timeout: Option<Duration>,
     ) -> Self {
         Self {
-            requester: Requester::new(ntx),
+            requester: Requester::new(ntx, broadcast_timeout),
             info_rx,
             warn_rx,
             event_rx,
@@ -46,11 +49,15 @@ impl Client {
 #[derive(Debug, Clone)]
 pub struct Requester {
     ntx: UnboundedSender<ClientMessage>,
+    broadcast_timeout: Option<Duration>,
 }
 
 impl Requester {
-    fn new(ntx: UnboundedSender<ClientMessage>) -> Self {
-        Self { ntx }
+    fn new(ntx: UnboundedSender<ClientMessage>, broadcast_timeout: Option<Duration>) -> Self {
+        Self {
+            ntx,
+            broadcast_timeout,
+        }
     }
 
     /// Tell the node to shut down.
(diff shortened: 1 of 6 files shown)


rustaceanrob:
This is by design. There is a possibility the first peer simply drops the `inv` announcement the first time we attempt to broadcast a transaction. It's the client responsibility to determine if they have already broadcast a transaction, they still want to try rebroadcasting, and to add timeouts accordingly. It would make sense to add a `Builder` method or direct argument on `broadcast_tx` to timeout.

randomlogin:
I think this PR may be closed, it was just a surprising behaviour to me in comparison to how other things broadcast txs, now it seems it is a valid design decision. 

Not sure about builder method or explicit timeout argument, is it needed or not?

rustaceanrob:
The only way to evaluate if a transaction is in a miner's memory pool is by heuristics, as we don't know if the transaction propagates beyond our peers. I think the intuition that we can improve broadcasting is correct. For instance, in Bitcoin Core's private broadcast feature, a transaction is advertised to peer A, and we wait for any other connection to broadcast the `Wtxid` back to us. This is a strong indication the transaction has propagated. The problem here is that this crate allows for one connection, so the logic would be different depending on the connections. I'd like to get #532 sorted first, but a combination of better heuristics and a timeout would be an improvement.

src/error.rs
```diff
@@ -28,6 +28,8 @@ pub enum ClientError {
     SendError,
     /// A channel was dropped before sending its value back.
     RecvError,
+    /// No peer requested the transaction within the configured broadcast timeout.
+    BroadcastTimeout,
```
rustaceanrob, line 32: Can we make a new error variant for this particular method (see `FetchBlockError`)

tests/core.rs
```diff
@@ -633,6 +633,98 @@ async fn tx_can_broadcast() {
         .unwrap();
 }
 
+// Verify that broadcasting a transaction that the peer already has.
+#[tokio::test]
+async fn tx_broadcast_does_not_hang_when_peer_has_tx() {
```
rustaceanrob, line 638: Can we just use the existing transaction broadcasting test instead of creating a new one?

src/client.rs
```diff
@@ -46,11 +49,15 @@ impl Client {
 #[derive(Debug, Clone)]
 pub struct Requester {
     ntx: UnboundedSender<ClientMessage>,
+    broadcast_timeout: Option<Duration>,
 }
 
 impl Requester {
-    fn new(ntx: UnboundedSender<ClientMessage>) -> Self {
-        Self { ntx }
+    fn new(ntx: UnboundedSender<ClientMessage>, broadcast_timeout: Option<Duration>) -> Self {
```
rustaceanrob, line 56: I'm not a fan of having this associated with the `Requester`. If a user would like to manage their own timeout, they should use `broadcast_tx`. Otherwise they should use a new method on the requester called `broadcast_tx_with_timeout`

randomlogin:
Do you still think this change is needed? An additional method and error for timeout? 
I find valid the current design where the caller decides how to handle the timeout.

rustaceanrob:
I'd like to leave this up as a reference if you don't mind. I am still undecided on if there should be a first-class API for the timeout, but in any case it should occur after I get to package relay.

PR #559, opened 2026-03-20
Don't clear filter queue on new block hash

### LINKED ISSUE #558: Peer ban during CF sync
Currently when we wait for compact filters if we receive an `inv` message about new block hash we discard the current filters request. 
When this discarded request arrives, we treat it as unexpected and think that the peer has misbehaved, which results in a peer's ban.

This happened to me on regtest, where the only peer got banned and the node has died, but it also happens on live networks, for example during the initial sync from an old block (when we have several batch requests of filters). This would ban peer and exhaust peer list which will lead to DNS peer discovery. 

If I understand it right, it is a high severity security problem, because it can be exploited by an attacker: we don't really check who has sent the `inv` message, which means an attacker can just send a garbage block hash (the attacker will also get eventually banned for that), which would result in banning a legitimate peer from which we had requested filters.

More precisely: 

When we sync cf headers, we check if there is a filter request made previously. If it has been cleared
src/chain/chain.rs:182:
```
174     pub(crate) fn sync_cf_headers(
...
178     ) -> Result<CFHeaderChanges, CFHeaderSyncError> {
179         let batch: CFHeaderBatch = cf_headers.into();                                                                                                                                                                                  
180         let request = self
181             .request_state
182             .last_filter_header_request
183             .ok_or(CFHeaderSyncError::UnexpectedCFHeaderMessage)?;
```

the call returns an Err which results in a disconnect and ban of the peer:

src/node.rs/495:
```
  474     async fn handle_cf_headers(
...
  478     ) -> Option<MainThreadMessage> {
  479         self.chain.send_chain_update().await;
  480         match self.chain.sync_cf_headers(peer_id, cf_headers) {
  481             Ok(potential_message) => match potential_message {
...
  490             },
  491             Err(e) => {
  492                 self.dialog.send_warning(Warning::UnexpectedSyncError {
  493                     warning: format!("Compact filter header syncing encountered an error: {e}"),
  494                 });
  495                 self.peer_map.ban(peer_id).await;
  496                 Some(MainThreadMessage::Disconnect)
  497             }    
```

When do we clear `last_filter_header_request`? 
It happens in src/chain/chain.rs:391, in `clear_compact_filter_queue`:
```
   388     // Reset the compact filter queue because we received a new block                                                                                                                                                                  
   389     pub(crate) fn clear_compact_filter_queue(&mut self) {
   390         self.request_state.agreement_state.reset_agreements();
   391         self.request_state.last_filter_header_request = None;
   392         self.request_state.pending_batch = None;
   393     }
```

Then `clear_compact_filter_queue` is called in `handle_inventory_blocks`, which is invoked when we receive "inv" gossip message.

src/node.rs:621
```
  597     async fn handle_inventory_blocks(
...
  601     ) -> Option<MainThreadMessage> {
  602         for block in blocks.iter() {
...
  607         }
  608         match self.state {
  609             NodeState::Behind => None,
  610             _ => {
  611                 if blocks
  612                     .into_iter()
  613                     .any(|block| !self.chain.header_chain.contains(block))
  614                 {
  615                     self.state = NodeState::Behind;
...
  621                     self.chain.clear_compact_filter_queue();                                                                                                                                                                            
  622                     Some(MainThreadMessage::GetHeaders(next_headers))
  623                 } else {
  624                     None
...
  628     }     
```

Potential mitigations:

1. Do not ban peer on `CFHeaderSyncError::UnexpectedCFHeaderMessage`. 
In that case we would just ignore compact filter messages. We could be spammed by these compact filters messages, but it seems to me it is expensive to do so; moreover just ignoring messages should be completely okay.

2. Do not clear the filter queue on "inv" messages at all. The fact that we received a new block doesn't necessarily mean that the filters returned are no longer valid. It speeds up the sync, as we don't discard previous results. In case of reorganization the queue still gets cleared (chain/chain.rs:137).

Also is it okay that we set the state to `Behind` whenever we receive an unknown block hash? Attacker can force us to re-request headers (seems to be fine).

randomlogin:
Removed clearing of the filter queue on new block hash.
Added a test which captures the previous behaviour of banning a legitimate peer on an "inv" message.

Fixes https://github.com/2140-dev/kyoto/issues/558 (partly? as the state still goes into `Behind` on new "inv" message received)

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -198,6 +198,12 @@ impl Chain {
             }
         }
         if request.stop_hash.ne(&batch.stop_hash()) {
+            if self
+                .request_state
+                .was_recently_requested(&batch.stop_hash())
+            {
+                return Ok(CFHeaderChanges::AddedToQueue);
+            }
             return Err(CFHeaderSyncError::UnrequestedStophash);
         }
         // Check that the start height we requested and the length of the batch are aligned.
@@ -291,6 +297,7 @@ impl Chain {
             .header_chain
             .block_hash_at_height(stop_hash_index)
             .unwrap_or(self.header_chain.tip_hash());
+        self.request_state.push_stop_hash(stop_hash);
         self.request_state.last_filter_header_request = Some(FilterHeaderRequest {
             expected_prev_filter_header: prev_header,
             start_height: last_unchecked_cfheader,
@@ -822,7 +829,7 @@ mod tests {
             filter_hashes: stale_hashes,
         };
         let cf_header_sync_res = chain.sync_cf_headers(0.into(), cf_headers);
-        assert!(cf_header_sync_res.is_err());
+        assert!(cf_header_sync_res.is_ok());
         let cf_headers = CFHeaders {
             filter_type: 0x00,
             stop_hash: scenario.last_block_hash(),
@@ -1062,7 +1069,7 @@ mod tests {
             filter_hashes: scenario.n_most_work_filter_hashes(4),
         };
         let cf_header_sync_res = chain.sync_cf_headers(1.into(), cf_headers);
-        assert!(cf_header_sync_res.is_err());
+        assert!(cf_header_sync_res.is_ok());
         chain.next_cf_header_message();
         let cf_headers = CFHeaders {
             filter_type: 0x00,
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -11,7 +11,7 @@ pub mod checkpoints;
 pub(crate) mod error;
 pub(crate) mod graph;
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 
 use bitcoin::constants::SUBSIDY_HALVING_INTERVAL;
 use bitcoin::hashes::{sha256d, Hash};
@@ -23,6 +23,8 @@ use bitcoin::{
 
 use crate::network::PeerId;
 use crate::HeaderCheckpoint;
+
+const MAX_PREV_STOP_HASHES: usize = 3;
 
 type Height = u32;
 
@@ -104,6 +106,7 @@ pub(crate) struct FilterRequestState {
     pub last_filter_header_request: Option<FilterHeaderRequest>,
     pub pending_batch: Option<(PeerId, CFHeaderBatch)>,
     pub agreement_state: FilterHeaderAgreements,
+    prev_cf_header_stop_hashes: VecDeque<BlockHash>,
 }
 
 impl FilterRequestState {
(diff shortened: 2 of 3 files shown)


tests/core.rs
```diff
@@ -633,6 +633,71 @@ async fn tx_can_broadcast() {
...
+    // The inv messages race with in-flight CF header responses.
+    for i in 0..20 {
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        rpc.generate_to_address(1, &miner).unwrap();
+        println!("Mined block during sync ({}/20)", i + 1);
+    }
+    let best = best_hash(rpc);
+    // The node should sync to the tip. If it banned bitcoind (its only peer on regtest),
+    // it dies with NoReachablePeers and never reaches FiltersSynced.
+    let sync_result = tokio::time::timeout(Duration::from_secs(30), async {
```
rustaceanrob: This is duplicate logic from `sync_assert` I think. You can add a timeout around that function

randomlogin: In the test we receive several FilterSynced events, assert_sync will fail if the sync hash differs from the best one.

randomlogin:
Going further: when we don't clear the last request on "inv" message, we might have a previous cf header request in flight, then we receive the new block header (and update the chain tip), request a new cf header batch and then the previous (stale) response arrives, so we treat it as unrequested, which still results in a peer getting banned.

src/chain/chain.rs:200
```
200        if request.stop_hash.ne(&batch.stop_hash()) {
201            return Err(CFHeaderSyncError::UnrequestedStophash);
202       }
```

Proposed solution: to have a cache of sent requests' `stop_hash`es, if the arriving response matches them, then we indeed requested it and we should not ban the peer, but merely ignore this response (as there is a new request already sent). 

There is a default count of 20 previous `stop_hash`es stored. On reorg the cache is cleared.

Please note, that there was a test `test_inv_no_queue` which captured this behaviour (that the stale response should be rejected), I modified that test to succeed. There might was another reason for this test and I might have broken it.

rustaceanrob:
At first glance I think this makes sense, I will have a more thorough review next week. I am traveling at the moment.

src/chain/chain.rs
```diff
@@ -390,6 +397,7 @@ impl Chain {
         self.request_state.agreement_state.reset_agreements();
         self.request_state.last_filter_header_request = None;
         self.request_state.pending_batch = None;
+        self.request_state.prev_cf_header_stop_hashes.clear();
```
rustaceanrob: Why clear these here? Shouldn't we hang on to them when making subsequent requests?

randomlogin: You are right, they should not be cleared on reorgs.

rustaceanrob:
Made some updates and squashed to a single commit. I was unable to reproduce the test locally without hitting timeouts. Can you verify this commit resolves your issue?

randomlogin:
Yes, it fixes the problem. 
It is weird that you couldn't reproduce, have you tried the (now-deleted) test with mining more blocks, say 10k? 

Also seems the cache capacity of 3 is enough for now, but if in future the synchronization becomes parallel it should be increased.

rustaceanrob:
> Also seems the cache capacity of 3 is enough for now, but if in future the synchronization becomes parallel it should be increased.

Yeah, correct.

> have you tried the (now-deleted) test with mining more blocks, say 10k?

I tried with less and more. It's likely a local issue but I was able to pass all other tests, so not sure where to start. For this one in particular the unit tests are sufficient.

PR #563, opened 2026-04-09
feat: add get_header(height) to Requester API

0xsiddharthks:
## Summary

- Exposes the existing `BlockTree::header_at_height` through the `Requester -> ClientMessage -> Node` channel
- Returns an `Option<IndexedHeader>` for a given block height — zero-cost local `BTreeMap` lookup, no network fetch
- Follows the same `ClientRequest<D, U>` + oneshot pattern used by `chain_tip()`, `peer_info()`, etc.

## Motivation

Consumers that maintain their own view of the chain (e.g., a bridge verifying deposit transactions) need to look up headers at specific heights to cross-check block data received from other sources. Without this, they'd have to maintain a parallel header index or make separate RPC calls to a full node.

## Testing

Extends the existing `various_client_methods` integration test to exercise `get_header`:
- Retrieves a header at a known height (1) and asserts the height matches
- Retrieves the header at the chain tip and asserts the block hash matches `chain_tip().hash`
- Requests a height beyond the chain and asserts `None` is returned

## Real-world usage

This feature is actively used in [Hashi](https://www.sui.io/hashi), a decentralized Bitcoin collateralization primitive on [Sui](https://www.sui.io/), built by [Mysten Labs](https://www.mystenlabs.com/). Specifically, `get_header` is used for security-critical verification that bitcoind is not lying about which block contains a deposit transaction — the bridge cross-checks headers at specific heights against its locally synced chain.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -7,6 +7,7 @@ use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::oneshot;
 
 use crate::chain::block_subsidy;
+use crate::chain::IndexedHeader;
 use crate::messages::ClientRequest;
 use crate::{Event, HeaderCheckpoint, Info, TrustedPeer, Warning};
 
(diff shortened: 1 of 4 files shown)


rustaceanrob:
LGTM ACK 4ce805f9ea991d368d2dc62426d3e8106dce13ba

PR #564, opened 2026-04-09
feat: add height_of_hash to Requester API

0xsiddharthks:
## Summary

- Exposes `BlockTree::height_of_hash` through the `Requester -> ClientMessage -> Node` channel
- Given a `BlockHash`, returns `Option<u32>` — the height of that block in the chain of most work, or `None` if the hash is unknown
- Zero-cost local `HashMap` lookup, no network fetch

## Motivation

Consumers that receive block hashes from external sources (e.g., a bridge watching for deposit confirmations) need to verify chain membership and determine block height without making separate RPC calls to a full node. This complements the existing `chain_tip()` API by allowing hash-to-height lookups against the locally synced header chain.

## Testing

Extends the existing `various_client_methods` integration test to exercise `height_of_hash`:
- Looks up the chain tip hash and asserts the returned height matches `chain_tip().height`
- Looks up an all-zeros hash and asserts `None` is returned

## Real-world usage

This feature is actively used in [Hashi](https://www.sui.io/hashi), a decentralized Bitcoin collateralization primitive on [Sui](https://www.sui.io/), built by [Mysten Labs](https://www.mystenlabs.com/). `height_of_hash` is used to verify block chain membership — given a block hash from bitcoind, Hashi can confirm it exists in the locally synced chain of most work without needing a separate RPC call.

### DIFF
--- a/src/chain/graph.rs
+++ b/src/chain/graph.rs
@@ -341,6 +341,14 @@ impl BlockTree {
         self.headers.get(hash).map(|node| node.header)
     }
 
+    // Returns the height of `hash` only when it sits on the canonical chain of most work.
+    // Returns `None` for unknown hashes and for hashes on stale/reorganized branches.
+    pub(crate) fn height_of_hash_canonical_only(&self, hash: BlockHash) -> Option<Height> {
+        let height = self.headers.get(&hash).map(|node| node.height)?;
+        let canonical = self.block_hash_at_height(height)?;
+        (canonical == hash).then_some(height)
+    }
+
     pub(crate) fn height_of_hash(&self, hash: BlockHash) -> Option<Height> {
         self.headers.get(&hash).map(|node| node.height)
     }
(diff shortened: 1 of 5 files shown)


rustaceanrob:
Needs rebase

rustaceanrob:
A few comments:
1. I realized there is a misleading code comment I wrote, as `height_of_hash` will return the height of a hash even if it belongs to a chain that was reorganized.
2. Are you simply polling if this hash is in the best chain, or do you need the associated height as well

rustaceanrob:
I am planning for a release in around the next week. Please provide feedback to those questions if you would like this included.

0xsiddharthks:
@rustaceanrob 
Thanks for the review.

For your questions:

1. Yeah, you're right. `BlockTree::height_of_hash` returns a height for any header in the graph, including orphaned branches, as it's reading the headers directly rather than cross-checking against `canonical_hashes`.
and yeah i agree that it's a bit misleading, and not actually what I want the public API to do.

2. We beed both — membership in the best chain *and* the height.
The process we're following is something like this:
- bitcoind reports a block hash for a deposit transaction; we look it up via `requester.height_of_hash(block_hash)`.
    - If it returns `None`, we treat that as "bitcoind reported a hash that isn't in kyoto's independently-validated chain of most work" (possibly malicious / lagging / on a stale branch) and skip the deposit.
    - If it returns `Some(height)`, we use that height to count confirmations via `(tip.height + 1) - height`.
- So "is this hash in the best chain, and if so at what height?" is the exact semantic we want.

I pushed `0576eff` (`fix: restrict height_of_hash to the canonical chain`) which tightens the lookup.

```rust
  pub(crate) fn height_of_hash(&self, hash: BlockHash) -> Option<Height> {
      let height = self.headers.get(&hash).map(|node| node.height)?;
      let canonical = self.block_hash_at_height(height)?;
      (canonical == hash).then_some(height)
  }
```

Now Some(h) is only returned when the hash is canonical at that height.
orphaned/reorg'd hashes return None. 

I extended test_depth_one_reorg to assert this: after a 1-block reorg, height_of_hash(old_tip) is None while height_of_hash(new_tip) returns the new height.

I also checked the two internal callers in `src/chain/chain.rs`
- `height_of_hash(batch.stop_hash()` at the CF header batch check
- `height_of_hash(filter_message.block_hash` in the filter check

Both operate on hashes tied to requests the node itself initiated for canonical work, so returning None for a non-canonical hash (e.g. during a reorg mid-sync) is the correct failure mode.

So i think with this update, we should be good.

Also, I'm happy to split the semantic fix into its own PR if you'd prefer to land the API addition and the semantic fix separately

rustaceanrob:
> I also checked the two internal callers 

I agree this should not be an issue, but to avoid unintended changes in sync behavior, I opted to make this a separate method on `BlockTree`. Also, I squashed to a single commit and signed it. Please note that this repository cannot merge unsigned commits as a policy, I noticed the last commit was unsigned.

PR #565, opened 2026-04-09
feat: add whitelist_only mode to restrict peer discovery

0xsiddharthks:
## Summary

- Adds `Builder::whitelist_only()` to restrict the node to only connect to explicitly configured peers
- When enabled, DNS seeding and addr gossip fallback are skipped entirely
- If all whitelist peers disconnect, `next_peer()` returns `None` and the node exits with `NoReachablePeers`, letting the caller restart with a fresh whitelist
- The node also skips sending `GetAddr` messages to connected peers

## Motivation

In controlled environments (private networks, infrastructure with known Bitcoin nodes, testing setups), consumers need strict control over which peers the node connects to. Previously, even with a whitelist configured, the node would fall back to DNS seeding or addr gossip when whitelist peers were exhausted. This flag makes peer selection deterministic and prevents any unintended connections to unknown peers.

## Testing

Adds two new integration tests:
- `whitelist_only_sync`: Builds a node with `whitelist_only()` and a single whitelisted bitcoind peer, verifies sync completes successfully
- `whitelist_only_no_peers_exits`: Builds a node with `whitelist_only()` and **no peers configured**, verifies the node exits with an error instead of falling back to DNS seeding — this tests the core restriction behavior

## Real-world usage

This feature is actively used in [Hashi](https://www.sui.io/hashi), a decentralized Bitcoin collateralization primitive on [Sui](https://www.sui.io/), built by [Mysten Labs](https://www.mystenlabs.com/). Hashi runs in controlled infrastructure where it must only connect to known, trusted Bitcoin peers — `whitelist_only` ensures no unexpected connections are made via DNS seeding or addr gossip.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,6 +342,7 @@ enum NodeState {
 struct Config {
     required_peers: u8,
     white_list: Vec<TrustedPeer>,
+    whitelist_only: bool,
     data_path: Option<PathBuf>,
     chain_state: Option<ChainState>,
     connection_type: ConnectionType,
@@ -355,6 +356,7 @@ impl Default for Config {
         Self {
             required_peers: 1,
             white_list: Default::default(),
+            whitelist_only: false,
             data_path: Default::default(),
             chain_state: Default::default(),
             connection_type: Default::default(),
--- a/src/node.rs
+++ b/src/node.rs
@@ -59,6 +59,7 @@ pub struct Node {
     chain: Chain,
     peer_map: PeerMap,
     required_peers: PeerRequirement,
+    whitelist_only: bool,
     dialog: Arc<Dialog>,
     block_queue: BlockQueue,
     client_recv: UnboundedReceiver<ClientMessage>,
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -355,6 +356,7 @@ impl Default for Config {
         Self {
             required_peers: 1,
             white_list: Default::default(),
+            whitelist_only: false,
```
rustaceanrob, line 359: `Default::default()` may be used here

src/node.rs
```diff
@@ -59,6 +59,7 @@ pub struct Node {
     chain: Chain,
     peer_map: PeerMap,
     required_peers: PeerRequirement,
+    whitelist_only: bool,
```
rustaceanrob, line 62: This field is not necessary because it is public on `PeerMap`. Please access it via that struct

rustaceanrob (commented):
Left a few comments

rustaceanrob:
Addressed comments and pushed to #567

PR #573, opened 2026-04-20
Add support for package relay

rustaceanrob:
Closes #532 

Adds type-safe support for relaying packages. 

Currently, the only supported package type is 1P1C, so this leveraged when sharing the transactions. First the child transaction is shared by `Wtxid`, which includes inputs that will be unknown to the receiver. The parent is never shared explicitly, instead we expect the receiver should request the inputs by their previous outpoint. The outpoint uses the legacy `Txid`, so parent transaction data is stored in a second map, indexed on `Txid`. Packages are confirmed only once the _parent_ has been shared. In the case of a single transaction, the flow of data does not change.

### DIFF
--- a/src/broadcaster.rs
+++ b/src/broadcaster.rs
@@ -1,40 +1,94 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
-use bitcoin::{Transaction, Wtxid};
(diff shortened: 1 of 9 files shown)


rustaceanrob:
Tested this works with a [CPFP](https://github.com/crustaceanagent/cpfp/tree/master) script.

Issue #561, opened 2026-04-08
Reorganization handling


randomlogin:
On regtest I had problems with syncing when reorganization occurs and started to investigate the problem.
What happened is that kyoto bans its only peer when reorg happens and I initially though that peer sends filters from reorged blocks => kyoto thinks these are garbage filters and bans the peer. 

Well, this could happen, but even after trying to fix this problem, (by introducing a cache of reorganized blocks to compare received filters against them) I couldn't sync properly.

Spend quite a lot of time debugging and I've realized there are several issues involved. 

1) The first one is the above mentioned, but it does **not** happen on regtest with a single peer. Still it is relevant for live networks (when we get filters and headers from different peers, which can occur, right?).

2) Another problem during sync on a reorganized chain: 
- we don't receive stale filters (as filters are sent one by one and if the peer made reorg, he would send the new filters)
- during reorganization we discard filter headers corresponding to reorganized blocks and when filters arrive we cannot process them (UnknownFilterHash).

If we do receive filter headers before the filters, we're fine and sync successfully. 

Setup to reproduce: 

1. Mine 4200 blocks (so it takes several filter requests to sync)
2. Make kyoto start syncing.
3. Invalidate a block. 
4. Mine two new blocks.

Please note that it is a race condition and it depends on your computer and you might need to invoke test several times to reproduce or to initially mine more/less blocks.

Logs with additional debugging prints:
```
In sync filter got message CFilter { filter_type: 0, block_hash: 2720ed35e8255268f42e5dad8705aa52e034fc8fd2bc2212c405700595ed6774, filter: [1, 106, 190, 0] }
banning the peer ManagedPeer { record: Record { addr: Ipv4(127.0.0.1), port: 44277, source: SourceId([182, 151, 35, 72]), services: ServiceFlags(3145), failed_attempts: 0, last_connection: None, last_attempt: None }, broadcast_min: FeeRate(250), ptx: Sender { chan: Tx { inner: Chan { tx: Tx { block_tail: 0x73696401a080, tail_position: 20 }, semaphore: Semaphore { semaphore: Semaphore { permits: 32 }, bound: 32 }, rx_waker: AtomicWaker, tx_count: 1, rx_fields: "..." } } }, handle: JoinHandle { id: Id(3) } }
[...]

PR #587, opened 2026-06-10
fix(net): request headers when a peer announces blocks via inv

0xsiddharthks:
Kyoto drops every `inv` message. With BIP-130 that's usually fine, but Core still falls back to announcing only the tip via `inv` when more than 8 blocks (`MAX_BLOCKS_TO_ANNOUNCE`) connect in one announcement round, or when a queued announcement gets reorged away. When that happens kyoto hears nothing until the 30-minute stale-tip check tears down every connection. On signet, which mines in bursts (we've seen 12-block runs), we counted 41 of these deaf episodes in 48 hours with no real 30-minute gap on chain.

This restores a minimal version of the inv handling removed in #570: an unknown block hash in an inv triggers a single `getheaders` to the announcing peer, and the response drives everything through the normal `handle_headers` path. Unlike the old handler, this one does not:
- mutate `NodeState`
- touch filter request state (the ban vector in #558)
- assume a new tip from the inv (the phantom-tip problem in #561)
- reset the stale-tip monitor (only the resulting `headers` response does)

Invs are ignored while a header sync is in flight or when every hash is already known, the probe only goes to the announcing peer, and the existing response timeout disconnects a peer that announces but can't deliver. A peer can already cause strictly more work today with unsolicited `headers`.

Tests: a reader unit test pins the inv mapping deterministically; the integration test burst-mines past the 8-block limit as a liveness check (a fast regtest box often stays under 8 per round, so the unit test is the regression anchor).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+## Fixed
+
+- Block announcements via `inv` trigger a `getheaders` request instead of being ignored
+
 ## 0.6.0
 
 ## Changed
--- a/src/network/mod.rs
+++ b/src/network/mod.rs
@@ -448,6 +448,7 @@ pub(crate) enum PeerMessage {
     FilterHeaders(CFHeaders),
     Filter(CFilter),
     Block(Block),
+    NewBlocks(Vec<BlockHash>),
     FeeFilter(FeeRate),
 }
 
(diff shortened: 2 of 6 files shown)


rustaceanrob:
Why is this in draft?

rustaceanrob:
Approach of c2977073df6e1f0a229bfdb9fcf182fc8fb750bd LGTM

The test is harder for me to reason about but I don't think this is particularly bad to add. I simplified it a bit in the head commit.

rustaceanrob (approved):
Unsure why this was in draft, and looks good to me otherwise. Going to merge this.

0xsiddharthks:
Hi @rustaceanrob , my bad, apologies for the delayed response. was afk last week.

Yeah the PR was ready!
Thanks again for reviewing and merging!

rustaceanrob:
I released `0.6.1` with this bug fixed

Issue #603, opened 2026-09-01
No correct `ScanType` for a brand-new wallet: checkpoints are too far back, and the tip needs a hash you can't get


thunderbiscuit:
I don't have an exact solution to propose because I know this has been explored and discussed in the past:

- #227 (with [comment from Rob here](https://github.com/2140-dev/kyoto/pull/227#issuecomment-2871637699))
- #456 

But I want to mention that as a user of the library, it feels like something is missing, since one of the basic use cases for the client is not well covered ("I just created a wallet. Let's go!" kind of thing). 

I myself again yesterday made the mistake of using `ScanType::Sync` and test it on mainnet, only to realize 20 minutes later that I was actually downloading all filters from genesis 😅. The error is bound to happen to many others IMO, because the API docs on the `ScanType::Recovery` are `Recover an existing wallet by scanning from the specified height.`, and the enum only offers you two choices: `Sync` and `Recovery`. You clearly shouldn't choose `Recovery` (or so you think!), since it's a never-seen-the-chain wallet, so you pick `Sync`. On a fresh wallet with no checkpoint, your sync will start from genesis, the very opposite of what you were trying to do.

### Idea 1: Hardcoded checkpoints
Kyoto used to have checkpoints in a checkpoint table, but they were removed in #456 with sensible reasonning:

> 1. Users will probably have to query a server for fees anyway. If they are building a new wallet for a user, they might as well take a more recent checkpoint to start the sync. By the time developers adopt the latest release, the checkpoints in the binary have gotten old.
> 2. Any developer supporting hardware wallets has no choice but to scan from Segwit or maybe Taproot
> 3. The checkpoints are only as useful as the frequency of releases, which will go down as the repository goes through refactoring

### Idea 2: Decouple filter headers and filters downloads
Kyoto currently doesn't download _all_ filter headers in the case of recovery with a Segwit or Taproot recovery checkpoint. But if I understand correctly, the filters start downloading at the recovery point, same as the filter headers. What if the API broke the two tasks and provided a third `ScanType` which would start downloading headers from the latest checkpoint (Taproot say), but only download filters starting from chain tip (or say 100 blocks back?). `ScanType::New` would be an obvious candidate name for the variant.

rustaceanrob:
Option 2 certainly seems possible with not too much effort. For option 1, I think release frequency will be the main problem here. The minute the checkpoints are released they become stale in a sense. I will have a look into option 2 soon

thunderbiscuit:
[...]

PR #605, opened 2026-09-14
feat: Limit initial scan to last N blocks

rustaceanrob:
"Closes" #603, although additional features similar to this one may be useful in the future, so I will leave that issue open.

This introduces a method on the builder so users can take a fixed `N` amount of filters from the tip, after they do headers sync. This should significantly speed up the startup time for new users that just want to interact with their wallet straight away.


thunderbiscuit (commented):
I don't feel like I'm knowledgable about the codebase enough to give a full ACK for this, but I do wanted to comment with 2 things:

1. I've been focused on trying to build an intuitive and flexible API for end users of compact block filters in production, and so that comes through the bdk_kyoto crate (for me at least). If this low-level change allows positive changes on that API, I'm all for it! But if I may, I'd suggest that it'd be great to see how it comes through for the end users on the bdk_kyoto crate before merging this (unless this PR is considered a strict improvement that can happen regardless of what bdk_kyoto does with it). A PR that would explore the separation between the checkpoint slash trusted block headers and the download of CBFs would be great, since this IMO needs more clarity and flexibility than the `ScanType` currently allows for/expresses.
2. When looking at the API, I was wondering what exactly happens if some of the configurations are not consistent with each other. It's a corner case but still; if a user provides a checkpoint that was merged an hour ago but sets `Builder::scan_filters_from_tip(100_000)`, it feels like a footgun since the behaviour there is underfined in my mental model of what the builder will do.

rustaceanrob:
> I'd suggest that it'd be great to see how it comes through for the end users on the bdk_kyoto crate

I drafted https://github.com/bitcoindevkit/bdk-kyoto/pull/165 which introduces a new type to enforce some of these constraints. There are three routes to start a sync, each one conveying what they are intended for. The builder is then constrained by switch route the user chooses. For instance, if a new wallet is the intention, the configuration introduced here is set under the hood. Other routes do not interact with this at all.

rustaceanrob:
#606 is preferred IMO

